PR #110, opened 2025-12-15
Add inverter model element

BrendanAnnable:


custom_components/haeo/flows/element.py
```diff
@@ -119,11 +119,9 @@ def _get_used_names(self) -> set[str]:
             if subentry.subentry_id != self._get_current_subentry_id()
         }
 
-    def _get_non_connection_element_names(self) -> list[str]:
```
TrentHouliston: If we want to be able to let users access the "raw graph" mode, we will still need something like this.

custom_components/haeo/translations/en.json
```diff
@@ -183,6 +138,38 @@
...
       "error": { "missing_name": "Name is required", "name_exists": "A device with this name already exists" }
     },
+    "inverter": {
+      "flow_title": "Inverter",
+      "entry_type": "Inverter",
+      "initiate_flow": { "user": "Inverter", "reconfigure": "Inverter" },
+      "step": {
+        "user": {
+          "title": "Add Inverter",
+          "description": "Set up an inverter with DC/AC coupling",
```
TrentHouliston: The DC/AC coupling in the translation probably will just confuse people who don't know and people who do know don't need it?

tests/model/test_scenarios.py
```diff
@@ -12,7 +11,7 @@ def test_simple_optimization() -> None:
     network.add("source_sink", "grid", is_source=True, is_sink=True)
     network.add("source_sink", "net", is_source=False, is_sink=False)
     network.add(
-        ELEMENT_TYPE_CONNECTION,
+        "connection",
```
TrentHouliston: This is one of the places that there should be a constant for from the modelling layer

custom_components/haeo/elements/inverter.py
```diff
@@ -0,0 +1,207 @@
...
+    inverter_outputs[INVERTER_DC_BUS_POWER_BALANCE] = dc_bus[SOURCE_SINK_POWER_BALANCE]
+
+    # source_target = DC to AC = EXPORT
+    # target_source = AC to DC = IMPORT
+    inverter_outputs[INVERTER_POWER_EXPORT] = replace(
+        connection[CONNECTION_POWER_SOURCE_TARGET], type=OUTPUT_TYPE_POWER
+    )
+    inverter_outputs[INVERTER_POWER_IMPORT] = replace(
+        connection[CONNECTION_POWER_TARGET_SOURCE], type=OUTPUT_TYPE_POWER
+    )
```
TrentHouliston: I think these are OUTPUT_FLOW types so no need for the replace.

Otherwise they will show up in the scenario plots as power being produced/consumed

custom_components/haeo/__init__.py
```diff
@@ -8,10 +8,14 @@
...
 from homeassistant.helpers import device_registry as dr
 
-from custom_components.haeo.const import CONF_ELEMENT_TYPE, CONF_NAME, ELEMENT_TYPE_NETWORK
+from custom_components.haeo.const import CONF_ELEMENT_TYPE, CONF_NAME
 
 from .coordinator import HaeoDataUpdateCoordinator
 
+# Network element type - defined here to avoid circular imports
+# (elements → schema → data → elements)
+ELEMENT_TYPE_NETWORK = "network"
```
TrentHouliston: Just define it the same as the rest of the element types!

custom_components/haeo/__init__.py
```diff
@@ -39,11 +43,15 @@ async def _ensure_network_subentry(hass: HomeAssistant, hub_entry: ConfigEntry)
     # Create Network subentry by adding it to the hub's subentries collection
     _LOGGER.info("Creating Network subentry for hub %s", hub_entry.entry_id)
 
+    # Use canonical name "network" for the network element
+    # This allows elements to connect to "network" consistently
+    network_name = "network"
```
TrentHouliston: I think it should inherit the name of the hub as the name of the network as the user hasn't explicitly provided a name for it otherwise and we can't as easily translate it

custom_components/haeo/elements/network.py
```diff
@@ -0,0 +1,130 @@
...
+def outputs(
+    name: str, model_outputs: Mapping[str, Mapping[ModelOutputName, OutputData]], _config: NetworkConfigData
+) -> Mapping[NetworkDeviceName, Mapping[NetworkOutputName, OutputData]]:
+    """Map model outputs to network-specific output names."""
+    network_node = model_outputs[name]
+
+    network_outputs: dict[NetworkOutputName, OutputData] = {}
+
+    # Power balance shadow price from the SourceSink node
+    network_outputs[NETWORK_POWER_BALANCE] = network_node[SOURCE_SINK_POWER_BALANCE]
```
TrentHouliston: I think the hub outputs should also be on the data here and we can collect them at the same time. Rather than making an extra path for them.

custom_components/haeo/flows/element.py
```diff
@@ -119,10 +122,25 @@ def _get_used_names(self) -> set[str]:
             if subentry.subentry_id != self._get_current_subentry_id()
         }
 
+    def _get_element_names(self) -> list[str]:
+        """Return all element names available as connection targets, excluding the current subentry.
+
+        This includes all elements (batteries, grids, inverters, network, etc.).
+        Network is now a proper element type, so it's included automatically.
```
TrentHouliston: This goes against the instructions to describe things as they are not referring to the things you changed. Also you can just use the _get_non_connection_element_names as you always did. You don't need to use this new function

tests/flows/test_element_flows.py
```diff
@@ -501,7 +472,7 @@ async def test_get_other_element_entries_filters_correctly(
     hub_entry: MockConfigEntry,
     flow_test_element_factory: FlowTestElementFactory,
 ) -> None:
-    """Verify participant filtering excludes non-endpoint subentries."""
+    """Verify participant filtering excludes non-element subentries like network."""
```
TrentHouliston: The test body doesn't agree with this description

tests/scenarios/test_scenarios.py
```diff
@@ -75,12 +75,15 @@ async def test_scenarios(
         await hass.async_block_till_done()
 
         # Create hub config entry and add to hass
+        # Use "network" as the hub title since scenario configs connect elements to "network"
+        # The network subentry is auto-created with the hub title as its name
         scenario_config = scenario_data["config"]
         mock_config_entry = MockConfigEntry(
             domain=DOMAIN,
+            title="network",  # Network name derives from hub title
```
TrentHouliston: Use the name Network

TrentHouliston (changes requested):
Finished reivew

TrentHouliston (approved):

PR #122, opened 2025-12-18
Change the way sensors are configured to allow runtime configuration

TrentHouliston:
This pull request introduces support for configurable number and switch entities for element parameters in the HAEO Home Assistant integration. These entities allow users to adjust element parameters at runtime via the Home Assistant UI, supporting both user-driven and externally-driven (read-only) modes. The PR also documents the new patterns for element output names and updates platform registration to include the new entity types.

### DIFF
--- a/custom_components/haeo/entities/haeo_number.py
+++ b/custom_components/haeo/entities/haeo_number.py
@@ -255,10 +255,11 @@ class HaeoInputNumber(CoordinatorEntity[HaeoDataUpdateCoordinator], RestoreNumbe
         """Set the value.
 
         In Editable mode: stores the value and triggers coordinator refresh.
-        In Driven mode: ignored - value is controlled by coordinator.
+        In Driven mode: ignored - writes current state to revert UI.
         """
         if self._entity_mode == ConfigEntityMode.DRIVEN:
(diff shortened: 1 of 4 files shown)


custom_components/haeo/elements/inverter.py
```diff
@@ -93,14 +87,11 @@ class InverterConfigData(TypedDict):
...
-    efficiency_dc_to_ac: NotRequired[PercentageFieldData]
-    efficiency_ac_to_dc: NotRequired[PercentageFieldData]
+    efficiency_dc_to_ac: NotRequired[PercentageSensorFieldData]
+    efficiency_ac_to_dc: NotRequired[PercentageSensorFieldData]
 
 
-CONFIG_DEFAULTS: dict[str, Any] = {
-    "efficiency_dc_to_ac": 100.0,
-    "efficiency_ac_to_dc": 100.0,
-}
```
TrentHouliston: These should still exist as default inits for the number field

TrentHouliston (commented):
I think that the schema types need to be built by composition such as Input[PowerFieldSensor] which can set the direction as it constructs the type information. Likewise something similar should allow passing the number selector details rather than hard coding it

TrentHouliston:
Superceeded by #168

PR #130, opened 2025-12-18
Update scenario tests

BrendanAnnable:
This PR updates the scenario tests to use the newer names and types.


custom_components/haeo/elements/inverter.py
```diff
@@ -141,11 +141,12 @@ def outputs(
 
     # source_target = DC to AC (inverting)
     # target_source = AC to DC (rectifying)
+    # Use power_flow type to indicate these are internal flows, not primary element outputs
     inverter_outputs[INVERTER_POWER_DC_TO_AC] = replace(
-        connection[CONNECTION_POWER_SOURCE_TARGET], type=OUTPUT_TYPE_POWER
+        connection[CONNECTION_POWER_SOURCE_TARGET], type=OUTPUT_TYPE_POWER_FLOW
```
TrentHouliston: It was already power flow! No need for replace!

BrendanAnnable: Done!

PR #142, opened 2025-12-21
Swap to using explicit ordering connections to make better battery decisions

TrentHouliston:
This pull request introduces major enhancements to the battery modelling logic, enabling support for time-varying battery section capacities and dynamic energy redistribution between battery sections. The changes include the introduction of a new `BatteryBalanceConnection` model element, significant updates to how battery sections and their capacities are handled, and adjustments to cost/incentive calculations for more realistic modelling.

The main reason for this upgrade is that we can vary the percentages on each section. That will allow us to have things like a "minimum reserve power" in the battery that changes based on time.

As a side effect it gets rid of the "cost based" ordering which is more of a guideline to the optimiser and replaces it with actual constraints which will enforce the ordering.

### DIFF
--- a/custom_components/haeo/elements/battery.py
+++ b/custom_components/haeo/elements/battery.py
@@ -479,22 +479,19 @@ def outputs(
         elif section_key == "overcharge":
             result[BATTERY_DEVICE_OVERCHARGE] = section_device_outputs
 
+    # Map section keys to device keys
+    section_to_device: dict[str, BatteryDeviceName] = {
+        "undercharge": BATTERY_DEVICE_UNDERCHARGE,
+        "normal": BATTERY_DEVICE_NORMAL,
+        "overcharge": BATTERY_DEVICE_OVERCHARGE,
+    }
+
     # Add balance power outputs to each section device
     # power_down = energy flowing into this section from above
     # power_up = energy flowing out of this section to above
     for i, section_key in enumerate(section_names):
-        # Get device for this section
-        if section_key == "undercharge":
-            device_key = BATTERY_DEVICE_UNDERCHARGE
-        elif section_key == "normal":
-            device_key = BATTERY_DEVICE_NORMAL
-        elif section_key == "overcharge":
-            device_key = BATTERY_DEVICE_OVERCHARGE
-        else:
-            continue
-
-        if device_key not in result:
-            continue
+        # Get device for this section - all sections in section_names have corresponding devices
+        device_key = section_to_device[section_key]
 
         # Accumulate power_down and power_up from all adjacent balance connections
         power_down_values: list[float] | None = None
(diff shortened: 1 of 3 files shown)


TrentHouliston (commented):
Here's a description of the modelling I want to use. The key here is that these constraints are setting minimums and maximum transfers not hard limits. So no objective or cost changes, the other parts of the system set the objective. The energy down shouldn't send more than is set because the battery storage limit constraints should stop it. The limit for the upward constraint should be forced to happen as otherwise it will exceed the limit. The lower one can't go higher otherwise it will exceed the lower section max or upper section min. 

# Battery Layer Flow Model

## Objective / Purpose

- Enforce **power flow between two sections** of a battery layer: upper → lower (down) and lower → upper (up).  
- **Downward flow**: must transfer as much power as physically possible to the lower section (at least the missing energy, capped by upper availability).  
- **Upward flow**: allowed only if the lower section is over its new capacity; flow can be anything between 0 and the available excess.  
- **Slack variables**: handle unmet demand (downward) and unavailable excess (upward) to maintain LP feasibility.  
- No explicit objective is required if the goal is **feasibility**; optional objective could maximize flows or minimize unmet demand.  

---

## Variables

### Downward flow

- $P_\text{down} \ge 0$ — power flowing from upper → lower section  
- $S_\text{down} \ge 0$ — slack representing unmet demand in lower section  

### Upward flow

- $P_\text{up} \ge 0$ — power flowing from lower → upper section  
- $S_\text{up} \ge 0$ — slack representing “excess not available” to flow upward  

---

## Parameters

- $E_\text{lower}$ — current energy in lower section (variable or constant)  
- $C_\text{lower}$ — full capacity of lower section  
- $E_\text{upper}$ — available energy in upper section  
- $C_\text{new}$ — new capacity limit for lower section  

Intermediate quantities:

$$
D = C_\text{lower} - E_\text{lower} \quad (\text{missing energy in lower})
$$

$$
X = E_\text{lower} - C_\text{new} \quad (\text{excess energy for upward flow})
$$

---

## Constraints

### 1. Downward flow

$$
\begin{aligned}
P_\text{down} + S_\text{down} &= D && \text{(must satisfy missing lower energy)} \\
0 \le P_\text{down} &\le E_\text{upper} && \text{(cannot transfer more than upper availability)} \\
S_\text{down} &\ge D - E_\text{upper} && \text{(slack only covers unmet demand)} \\
S_\text{down} &\ge 0 && \text{(non-negativity)}
\end{aligned}
$$

**Behavior:**

- If upper has enough energy → $P_\text{down} = D$, $S_\text{down} = 0$  
- If upper has less → $P_\text{down} = E_\text{upper}$, $S_\text{down} = D - E_\text{upper}$  
- If lower is full → $P_\text{down} = 0$, $S_\text{down} = 0$  

---

### 2. Upward flow

$$
\begin{aligned}
P_\text{up} + S_\text{up} &= E_\text{lower} - C_\text{new} && \text{(slack captures unavailable excess)} \\
P_\text{up} &\ge 0 && \text{(cannot flow negative)} \\
S_\text{up} &\ge 0 && \text{(non-negativity)}
\end{aligned}
$$

**Behavior:**

- If $E_\text{lower} \le C_\text{new}$ → $P_\text{up} = 0$, $S_\text{up} = C_\text{new} - E_\text{lower}$  
- If $E_\text{lower} > C_\text{new}$ → $P_\text{up} \in [0, E_\text{lower} - C_\text{new}]$, $S_\text{up} = E_\text{lower} - C_\text{new} - P_\text{up}$  

---

### 3. Optional: Feasibility notes

- LP is fully feasible in all scenarios (partial/full layers, over/under capacity)  
- Slack variables ensure that the solver cannot pick infeasible negative flows  
- The model is purely linear and does not require MILP binaries unless exact priorities are needed  

---

### 4. Example Scenarios

| Scenario | $P_\text{down}$ | $S_\text{down}$ | $P_\text{up}$ | $S_\text{up}$ |
|----------|-----------------|-----------------|---------------|---------------|
| Lower full, upper empty | 0 | 0 | 0 | 0 |
| Lower partial, upper enough | D | 0 | 0 | 0 |
| Lower partial, upper limited | $E_\text{upper}$ | D-$E_\text{upper}$ | 0 | 0 |
| Lower > new capacity | any feasible | 0 | ≤ excess | slack = excess - P_up |
| Lower ≤ new capacity | 0 | 0 | 0 | slack = C_new - E_lower |

---

**Summary:**  

- **Downward flow:** enforces minimum transfer to fill lower section, capped by upper availability.  
- **Upward flow:** enforces maximum allowed transfer when lower section is over capacity, lower bound = 0.  
- Slack variables guarantee **LP feasibility** and handle variable-dependent energy levels.

PR #168, opened 2025-12-29
Separate the inputs to load and update independently

TrentHouliston:
Rather than doing it in a single pass, this makes it so the inputs for the optimsiation are loaded by their own coordinators/cycle, which is then used by the optimiser.

This way it is easier to update/change the inputs and we can make them runtime editable, then the optimiser just has to focus on loading and optimising them

### DIFF
--- a/custom_components/haeo/const.py
+++ b/custom_components/haeo/const.py
@@ -1,5 +1,6 @@
 """Constants for the Home Assistant Energy Optimizer integration."""
 
+import enum
 from typing import Final, Literal
 
 # Integration domain
@@ -76,3 +77,16 @@ type NetworkDeviceName = Literal["network"]
 NETWORK_DEVICE_NAMES: Final[frozenset[NetworkDeviceName]] = frozenset(
     (NETWORK_DEVICE_NETWORK := "network",),
 )
+
+
+class ConnectivityLevel(enum.StrEnum):
+    """Connectivity level for element types in connection selectors.
+
+    - ALWAYS: Always shown in connection selectors
+    - ADVANCED: Only shown when advanced mode is enabled
+    - NEVER: Never shown in connection selectors
+    """
+
+    ALWAYS = enum.auto()
+    ADVANCED = enum.auto()
+    NEVER = enum.auto()
--- a/custom_components/haeo/elements/__init__.py
+++ b/custom_components/haeo/elements/__init__.py
@@ -27,14 +27,19 @@ Sub-element Naming Convention:
 """
 
 from collections.abc import Mapping, Sequence
-import enum
 import logging
 from typing import Any, Final, Literal, NamedTuple, Protocol, TypeGuard, get_origin, get_type_hints, runtime_checkable
 
 from homeassistant.config_entries import ConfigEntry, ConfigSubentry
 from homeassistant.core import HomeAssistant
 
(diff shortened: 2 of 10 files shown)


custom_components/haeo/__init__.py
```diff
@@ -103,15 +124,17 @@ async def async_setup_entry(hass: HomeAssistant, entry: HaeoConfigEntry) -> bool
     # Ensure required subentries exist (auto-create if missing)
     await _ensure_required_subentries(hass, entry)
 
-    # Store coordinator in runtime data first (required for platform setup)
-    coordinator = HaeoDataUpdateCoordinator(hass, entry)
-    entry.runtime_data = coordinator
+    # Create network coordinator for optimization
+    network_coordinator = HaeoDataUpdateCoordinator(hass, entry)
```
TrentHouliston: Keeping it named coordinator throughout the code rather than network coordinator will reduce the diff and it's only named this way because early in this PR we split it

custom_components/haeo/__init__.py
```diff
@@ -103,15 +124,17 @@ async def async_setup_entry(hass: HomeAssistant, entry: HaeoConfigEntry) -> bool
...
+
+    # Store runtime data (required for platform setup)
+    entry.runtime_data = HaeoRuntimeData(network_coordinator=network_coordinator)
 
     # Register update listener for config changes and subentry additions/removals
     entry.async_on_unload(entry.add_update_listener(async_update_listener))
 
     # Trigger initial optimization on startup
-    await coordinator.async_config_entry_first_refresh()
+    await network_coordinator.async_config_entry_first_refresh()
```
TrentHouliston: I think this needs to come after setting up the input platforms as otherwise there won't be anything to optimise against. Or at least I think there is more nuance to how this setup should be ordered.

Perhaps rather than doing an optimisation straight away we should just wait for it to do it naturally due to the entities being triggered. I'm fairly sure this was only done this way to make the scenario integration tests easier.

That still requires the platforms being set up in the right order so we can have events that wait on them.

Or maybe I'm overthinking this and it's fine.

custom_components/haeo/entities/haeo_horizon.py
```diff
@@ -0,0 +1,166 @@
...
+        )
+
+    @callback
+    def _async_scheduled_update(self, _now: datetime) -> None:
+        """Handle scheduled update when period boundary is reached."""
+        self._update_horizon()
+        self.async_write_ha_state()
+
+        # Notify all subscribers
+        for subscriber in self._subscribers:
```
TrentHouliston: Won't this be handled by the write state? Anything watching this entity should update then.

custom_components/haeo/entities/haeo_horizon.py
```diff
@@ -0,0 +1,166 @@
...
+        if self._unsub_timer is not None:
+            self._unsub_timer()
+            self._unsub_timer = None
+        await super().async_will_remove_from_hass()
+
+    @callback
+    def async_subscribe(self, callback_fn: Callable[[], None]) -> Callable[[], None]:
+        """Subscribe to horizon updates.
+
+        Args:
```
TrentHouliston: Just in general I need to update the AI rules. I would prefer to not have the args and returns when they are self evident. Documentation is great, but when it's repeating what's already in the signature it just becomes brittle. If one is updated without the other it gets confusing and it didn't add much anyway.

PR #175, opened 2026-01-03
Simplify subentry configuration

BrendanAnnable:


custom_components/haeo/__init__.py
```diff
@@ -159,6 +160,75 @@ async def _ensure_required_subentries(hass: HomeAssistant, hub_entry: ConfigEntr
         _LOGGER.debug("Switchboard node created successfully")
 
 
+async def _migrate_constants_to_entity_refs(hass: HomeAssistant, entry: HaeoConfigEntry) -> None:
```
TrentHouliston: I don't think we need to have a migration, if we continue to store the actual float constants in the configuration the storage format doesn't change.

Also I think once we do start introducing migration flows they should be more structured than just putting them in the init

custom_components/haeo/__init__.py
```diff
@@ -181,6 +251,11 @@ async def async_update_listener(hass: HomeAssistant, entry: HaeoConfigEntry) ->
 
 async def async_setup_entry(hass: HomeAssistant, entry: HaeoConfigEntry) -> bool:
     """Set up Home Assistant Energy Optimizer from a config entry."""
+    # Set up configurable sentinel entities for each device_class if not already present
+    from custom_components.haeo.flows.constants import ensure_configurable_entities_exist  # noqa: PLC0415
```
TrentHouliston: Should be no reason to avoid the linting rule here. Just import it at the top

custom_components/haeo/elements/battery/flow.py
```diff
@@ -99,15 +77,37 @@ async def async_step_user(self, user_input: dict[str, Any] | None = None) -> Sub
...
+        default_name = translations.get(f"component.{DOMAIN}.config_subentries.{ELEMENT_TYPE}.flow_title", "Battery")
 
+        entity_metadata = extract_entity_metadata(self.hass)
+        exclusion_map = build_exclusion_map(INPUT_FIELDS, entity_metadata)
         participants = self._get_participant_names()
-        schema = _build_step1_schema(participants)
+        schema = _build_step1_schema(participants, exclusion_map)
 
-        # Apply default mode selections
-        defaults = get_mode_defaults(INPUT_FIELDS, BatteryConfigSchema)
```
TrentHouliston: Defaults should live in the schema without extra hard coding in the flow!

custom_components/haeo/elements/battery_section/flow.py
```diff
@@ -25,8 +35,13 @@ def _filter_incompatible_entities(
     return [v.entity_id for v in entity_metadata if not v.is_compatible_with(accepted_units)]
 
 
-def _build_schema(entity_metadata: list[EntityMetadata]) -> vol.Schema:
-    """Build the voluptuous schema for battery section configuration."""
+def _get_field(field_name: str) -> Any:
+    """Get field info by name."""
+    return next(f for f in INPUT_FIELDS if f.field_name == field_name)
```
TrentHouliston: I was actually intending on changing the INPUT_FIELDS to be a parallel dict to the typed dict so you can iterate over them together. That gets rid of a lot of this noise

custom_components/haeo/elements/battery_section/flow.py
```diff
@@ -69,21 +88,98 @@ async def async_step_user(self, user_input: dict[str, Any] | None = None) -> Sub
...
+
+        # Ensure configurable entity exists before building schema
+        ensure_configurable_entities_exist()
+
+        # Get default name from translations
+        translations = await async_get_translations(
+            self.hass, self.hass.config.language, "config_subentries", integrations=[DOMAIN]
+        )
+        default_name = translations.get(
+            f"component.{DOMAIN}.config_subentries.{ELEMENT_TYPE}.flow_title", "Battery section"
```
TrentHouliston: Don't have fallbacks, if we fail to get the translation we should error. Otherwise these sorts of bugs live silently for ages without being caught.

custom_components/haeo/elements/battery_section/flow.py
```diff
@@ -94,21 +190,90 @@ async def async_step_reconfigure(self, user_input: dict[str, Any] | None = None)
...
+
+        ensure_configurable_entities_exist()
+
+        entity_metadata = extract_entity_metadata(self.hass)
+        schema = _build_step1_schema(entity_metadata)
+
+        current_data = dict(subentry.data)
+        entity_defaults = get_entity_selection_defaults(INPUT_FIELDS, BatterySectionConfigSchema, current_data)
+        defaults: dict[str, Any] = dict(entity_defaults)
+        defaults[CONF_NAME] = subentry.data.get(CONF_NAME)
```
TrentHouliston: I feel like getting the translations could be done in their schemas for their defaults. One less thing for the flow to worry about.

custom_components/haeo/elements/battery_section/schema.py
```diff
@@ -7,13 +7,54 @@
...
+
 ELEMENT_TYPE: Final = "battery_section"
 
 # Configuration field names
 CONF_CAPACITY: Final = "capacity"
 CONF_INITIAL_CHARGE: Final = "initial_charge"
 
 
+# Input fields for entity selection
+INPUT_FIELDS: Final[tuple[InputFieldInfo[NumberEntityDescription], ...]] = (
```
TrentHouliston: Yeah this makes more sense as a dict for sure.

TrentHouliston (commented):
Partial review. Many of the comments apply to all flows not just the specific one I commented on.

custom_components/haeo/elements/inverter/schema.py
```diff
@@ -65,7 +59,7 @@
             native_step=0.1,
         ),
         output_type=OutputType.EFFICIENCY,
-        default=100.0,
```
TrentHouliston: I feel like we should be able to have a default for almost everything. Stored in one spot and just in the format of the storage schema.

custom_components/haeo/entities/haeo_number.py
```diff
@@ -27,17 +29,17 @@ class ConfigEntityMode(Enum):
     DRIVEN = "driven"  # Value driven by external entity
 
 
-class HaeoInputNumber(NumberEntity):
+class HaeoInputNumber(RestoreEntity, NumberEntity):
```
TrentHouliston: I had removed restore entity so that it would store and restore from the configuration rather than a separate store

TrentHouliston (commented):
More partial review

TrentHouliston (commented):
More partial review.

PR #182, opened 2026-01-04
Add configurable entity sentinel

BrendanAnnable:


custom_components/haeo/haeo.py
```diff
@@ -0,0 +1,35 @@
...
+from homeassistant.core import HomeAssistant
+from homeassistant.helpers.entity_platform import AddEntitiesCallback
+
+from custom_components.haeo.entities.sentinels import ConfigurableEntity
+
+# Sentinel entities are static and never update
+PARALLEL_UPDATES = 0
+
+
+async def async_setup_entry(
```
TrentHouliston: Setup entry implies that there is one of these per haeo hub, not per home assistant instance. You probably want just a single one of these, not one per hub.

PR #203, opened 2026-01-10
Add suggested optimisation control actions based on the outputs and shadow prices

TrentHouliston:
The first version of the automation docs are actually pretty bad with their recommendations when trying to interface with real hardware.

This tries to uplift them to better describe how to interpret the different outputs and what you should actually do


TrentHouliston:
Found a better way than what this was trying

PR #216, opened 2026-01-12
Unload the integration when we have a timeout error

TrentHouliston:
When the integration fails to load because of a timeout error (e.g. we're awaiting a very slow entity) we throw a retry error. That then makes it retry again later but doesn't unload the things it already loaded. Those things then throw errors because they are already configured creating a death spiral of reloads failing

### DIFF
--- a/custom_components/haeo/__init__.py
+++ b/custom_components/haeo/__init__.py
@@ -177,28 +177,41 @@ async def async_setup_entry(hass: HomeAssistant, entry: HaeoConfigEntry) -> bool
     # Register cleanup - will be called on failure or unload
     entry.async_on_unload(lambda: async_unload_sentinel_entities(hass))
 
-    # Start horizon manager's scheduled updates
-    horizon_manager.start()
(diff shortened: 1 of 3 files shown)


custom_components/haeo/__init__.py
```diff
@@ -164,46 +172,77 @@ async def async_setup_entry(hass: HomeAssistant, entry: HaeoConfigEntry) -> bool
...
     await async_setup_sentinel_entities(hass)
+    # Register cleanup - will be called on failure or unload
+    entry.async_on_unload(lambda: async_unload_sentinel_entities(hass))
 
     # Start horizon manager's scheduled updates
     horizon_manager.start()
-
-    # Register cleanup on unload
+    # Register cleanup - will be called on failure or unload
     entry.async_on_unload(horizon_manager.stop)
```
TrentHouliston: This should follow the pattern where the start returns the stop and you pass it directly into unload.

PR #227, opened 2026-01-14
Change connections to be made of composable components

TrentHouliston:
Changes how connections work to be composable connections.

Currently we have connections for specific purposes that do a few things, like limiting power and adding prices. Or doing battery balance.

Eventually I want to add things like power tagging, filtering, demand pricing, etc and all the combinations of these makes a nightmare of complexity.

So instead I'm refactoring connections to be composed of the things they're made of.

So you can make something like a Lossy Priced Limited connection, that will do efficiency, pricing, and power limits. That way we can combine all the things we want to do to power on the move by just composing the operations we want to apply to them.

### DIFF
--- a/tests/entities/test_haeo_number.py
+++ b/tests/entities/test_haeo_number.py
@@ -974,7 +974,7 @@ async def test_editable_mode_with_boundaries_field(
     values = entity.get_values()
     assert values is not None
     assert len(values) == 3  # All 3 boundary timestamps
-    assert all(v == 50.0 for v in values)
+    assert all(v == 0.5 for v in values)
 
 
 async def test_async_load_data_with_boundaries_field(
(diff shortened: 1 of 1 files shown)


custom_components/haeo/elements/connection/adapter.py
```diff
@@ -173,63 +190,101 @@ def inputs(self, config: Any) -> dict[str, InputFieldInfo[Any]]:
...
         """Map model outputs to connection-specific output names."""
         connection = model_outputs[name]
+        power_source_target = connection[CONNECTION_POWER_SOURCE_TARGET]
+        power_target_source = connection[CONNECTION_POWER_TARGET_SOURCE]
+        if not isinstance(power_source_target, OutputData):
+            msg = f"Expected OutputData for {name!r} {CONNECTION_POWER_SOURCE_TARGET}"
+            raise TypeError(msg)
+        if not isinstance(power_target_source, OutputData):
+            msg = f"Expected OutputData for {name!r} {CONNECTION_POWER_TARGET_SOURCE}"
+            raise TypeError(msg)
```
TrentHouliston: More of the typing errors

PR #301, opened 2026-02-08
Add lexicographic multiobjective optimization with calibrated blending

TrentHouliston:
Replace the single-objective cost minimization with a two-objective lexicographic framework that deterministically breaks ties among cost-equivalent solutions.

## Motivation

When multiple schedules share the same minimum cost, the LP solver previously returned an arbitrary vertex.
This caused non-deterministic behavior: battery charge/discharge schedules could flip between equivalent solutions across successive optimizations, producing unstable sensor outputs.

A secondary time-preference objective now selects among cost-equivalent solutions by preferring earlier energy transfers, yielding stable and physically intuitive schedules without affecting the optimal cost.

## Solve modes

Three solver modes are available via `SolveOptions`:

- **Calibrated** (default): Two-phase lex on the first call, then binary-searches for a blend weight that reproduces lex decision variables in a single weighted-sum solve. Subsequent calls use the blended fast path for efficient warm-starts.
- **Lex**: Three-phase lexicographic solve (minimize primary, minimize secondary with primary constrained, re-minimize primary with secondary constrained). Restores clean shadow prices at the cost of an extra solve.
- **Blended**: Single-solve weighted sum (`primary + weight × secondary`) with a user-specified blend weight.

## Time-preference ordering

Each connection contributes a secondary objective term: `priority × T + (t + 1)` weights energy transfer by both connection priority group and time step. Priorities are auto-computed from endpoint properties:

- Own power (solar, load) is preferred over external (grid import/export)
- Time-sensitive connections (loads, generation) are preferred over invariant ones
- Alphabetical tiebreaking by source/target/name ensures determinism

Connections with different priorities receive non-overlapping weight ranges, so the secondary objective cleanly separates priority groups from within-group time ordering.

### Model layer (`core/model/`)
- `network.py`: New `SolveOptions` discriminated union (`LexOptions`, `BlendedOptions`, `CalibratedOptions`) with `SimplexTuning` base. Lexicographic, blended, and calibrated solve implementations. Binary-search calibration of blend weight with safe-zone centering.
- `element.py`: `cost()` return type narrowed to `highs_linear_expression | None`
- `connection.py`: `cost()` returns `(primary, secondary)` tuple. New `is_external`, `is_time_sensitive` flags and `sort_key` property for deterministic priority. Time-preference secondary objective generation.
- `segment.py`: `cost()` uses `@cost` decorator with proper type narrowing
- `output_data.py`: New `priority` field for connection outputs

### Adapters (`core/adapters/elements/`)
- Grid: Sets `is_external=True` on import/export connections
- Load: Sets `is_time_sensitive=True` on load connections
- Solar: Sets `is_time_sensitive=True` on solar connections

### Integration
- `coordinator.py`: Threads `priority` through `CoordinatorOutput`
- `haeo_sensor.py`: Exposes `priority` as sensor attribute

### CI and testing
- Separate scenario test job in CI
- `pytest-benchmark` integration with benchmark marker
- New `test_benchmark.py` for solver performance testing
- Updated scenario snapshots for deterministic ordering
- HiGHS stubs updated for new API surface

### Documentation
- Modeling overview: Lexicographic solve process, secondary objective formulation
- Connection model docs: Time-preference cost contribution with math
- Developer guide: Updated `@cost` decorator docs, cost modeling section
- Architecture: Updated optimization description
- User guide: Lexicographic ordering explanation


custom_components/haeo/core/model/network.py
```diff
@@ -232,3 +351,76 @@ def constraints(self) -> dict[str, dict[str, highs_cons | list[highs_cons]]]:
...
+def _as_linear_expression(expression: highs_linear_expression | highs_var) -> highs_linear_expression:
+    """Coerce a variable to a linear expression when needed."""
+    if isinstance(expression, highs_linear_expression):
+        return expression
+    return Highs.qsum([expression])
+
+
+def _clear_linear_objectives(solver: Highs) -> None:
+    """Clear any multiobjective state if the API is available."""
+    cast("Any", solver).clearLinearObjectives()
```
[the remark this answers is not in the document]
TrentHouliston: :robot: This is not an issue — `clearLinearObjectives()` is declared in `stubs/highspy/__init__.pyi` (line 204) and confirmed to exist at runtime in our pinned highspy version. The reviewer's premise that the method "isn't declared in stubs" was incorrect. No guard needed.

custom_components/haeo/core/model/network.py
```diff
@@ -186,20 +195,130 @@ def optimize(self) -> float:
...
+
+        primary = objectives[0]
+        secondary = objectives[1] if len(objectives) > 1 else None
+
+        if primary is None:
+            self._relax_lex_constraint()
+            h.run()
+            _ensure_optimal(h)
+            return 0.0
+
```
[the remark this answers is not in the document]
TrentHouliston: :robot: Fixed. When `primary is None` but `secondary` exists, we now set the secondary cost vector before solving. This ensures time-preference ordering is still optimized in "no monetary cost" models, while still returning `0.0` as the primary cost value.

custom_components/haeo/core/model/element.py
```diff
@@ -255,14 +255,27 @@ def cost(self) -> Any:
...
+    combined: list[Any] = []
+    for index in range(max_len):
+        terms = [items[index] for items in objectives if len(items) > index]
+        if not terms:
+            continue
+        if len(terms) == 1:
+            combined.append(terms[0])
+        else:
+            combined.append(Highs.qsum(terms))
+    return combined
```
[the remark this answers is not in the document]
TrentHouliston: :robot: Fixed. The root cause was in the `cost()` methods of `element.py` and `segment.py`, which filtered `None` from cost value lists before passing to `_combine_objective_lists`: `[item for item in cost_value if item is not None]`. This collapsed `[None, secondary_expr]` into `[secondary_expr]`, shifting it to index 0.

The fix preserves `None` entries in cost value lists so positional semantics are maintained. Additionally, `_combine_objective_lists` is now consolidated into a single shared implementation in `element.py` (see thread 6).

custom_components/haeo/core/model/elements/segments/segment.py
```diff
@@ -179,15 +180,30 @@ def cost(self) -> Any:
...
+    combined: list[Any] = []
+    for index in range(max_len):
+        terms = [items[index] for items in objectives if len(items) > index]
+        if not terms:
+            continue
+        if len(terms) == 1:
+            combined.append(terms[0])
+        else:
+            combined.append(Highs.qsum(terms))
+    return combined
```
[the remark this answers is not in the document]
TrentHouliston: :robot: Fixed — same root cause as thread 2. The `Segment.cost()` method now preserves `None` entries in cost value lists. Additionally, the local `_combine_objective_lists` copy has been removed and replaced with an import from `element.py`.

custom_components/haeo/core/model/elements/connection.py
```diff
@@ -272,25 +275,50 @@ def constraints(self) -> dict[str, highs_cons | list[highs_cons]]:
...
+        # Place time preference at the secondary objective index
+        if (time_preference := self._time_preference_objective()) is not None:
+            # Pad with None so time preference lands at SECONDARY_OBJECTIVE_INDEX
+            while len(combined) < SECONDARY_OBJECTIVE_INDEX:
+                combined.append(None)
+
+            if len(combined) == SECONDARY_OBJECTIVE_INDEX:
+                combined.append(time_preference)
+            else:
+                combined[SECONDARY_OBJECTIVE_INDEX] = Highs.qsum([combined[SECONDARY_OBJECTIVE_INDEX], time_preference])
```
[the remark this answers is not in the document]
TrentHouliston: :robot: This was already fixed in a previous commit. The current code already has the `if existing is None` check:

```python
existing = combined[SECONDARY_OBJECTIVE_INDEX]
if existing is None:
    combined[SECONDARY_OBJECTIVE_INDEX] = time_preference
else:
    combined[SECONDARY_OBJECTIVE_INDEX] = Highs.qsum([existing, time_preference])
```

custom_components/haeo/core/model/element.py
```diff
@@ -255,14 +247,29 @@ def cost(self) -> Any:
             method = getattr(self, name)
             if (cost_value := method()) is not None:
                 if isinstance(cost_value, list):
-                    costs.extend(cost_value)
+                    costs.append([item for item in cost_value if item is not None])
```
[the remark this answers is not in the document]
TrentHouliston: :robot: Fixed. The `None`-filtering `[item for item in cost_value if item is not None]` has been removed from `cost()` in both `element.py` and `segment.py`. Cost value lists now preserve `None` entries so positional semantics are maintained through `_combine_objective_lists`.

custom_components/haeo/core/model/element.py
```diff
@@ -255,14 +247,29 @@ def cost(self) -> Any:
...
+    combined: list[highs_linear_expression | None] = []
+    for index in range(max_len):
+        terms = [items[index] for items in objectives if len(items) > index]
+        if not terms:
+            combined.append(None)
+        elif len(terms) == 1:
+            combined.append(terms[0])
+        else:
+            combined.append(Highs.qsum(terms))
+    return combined
```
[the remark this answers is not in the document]
TrentHouliston: :robot: Fixed. All four copies have been consolidated into a single `_combine_objective_lists` in `element.py`. The other three modules (`connection.py`, `segment.py`, `network.py`) now import it from there. The single implementation correctly handles `None` entries by filtering during combining rather than before.

custom_components/haeo/core/model/network.py
```diff
@@ -232,3 +350,71 @@ def constraints(self) -> dict[str, dict[str, highs_cons | list[highs_cons]]]:
...
+    per objective switch).
+    """
+    vectors: list[NDArray[np.float64]] = []
+    for obj in objectives:
+        vec = np.zeros(n_vars, dtype=np.float64)
+        if obj is not None:
+            idxs, vals = obj.unique_elements()
+            vec[idxs] = vals
+        vectors.append(vec)
+    return vectors
```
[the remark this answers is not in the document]
TrentHouliston: :robot: This is a theoretical concern that doesn't apply in practice. All cost expressions in this codebase are built from `price * power * periods` products — variable-coefficient terms with no constant offsets. The `highs_linear_expression` objects produced by these operations have structurally zero constants.

The `_set_cost_vector` call to `changeObjectiveOffset(0.0)` is a safety reset, not a lossy operation. If a constant were ever introduced in a cost expression, it would be a modeling error (constants in LP objectives don't affect the optimal solution, only the reported objective value). No code change needed.

custom_components/haeo/core/model/network.py
```diff
@@ -178,20 +258,249 @@ def optimize(self) -> float:
...
 
-        msg = f"Optimization failed with status: {h.modelStatusToString(status)}"
-        raise ValueError(msg)
+        h.changeObjectiveSense(ObjSense.kMinimize)
+
+        primary = objectives[0]
+        secondary = objectives[1] if len(objectives) > 1 else None
+
+        if primary is None:
+            self._relax_lex_constraint()
```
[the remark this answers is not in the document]
TrentHouliston: :robot: Fixed — same issue as thread 1. When `primary is None` but `secondary` exists, `optimize()` now sets the secondary cost vector before solving, ensuring the time-preference objective is still minimized while returning `0.0` as the primary cost.

docs/modeling/model-layer/connections/connection.md
```diff
@@ -87,7 +87,26 @@ Efficiency losses are applied inside the segment chain.
...
+$$
+
+Where $c$ is the connection index, $T$ is the number of periods, and $t$ is the time step.
+The time-preference objective for connection $c$ is:
+
+$$
+\sum_{t=0}^{T-1} w_{c,t} \cdot P_{\text{in}}(t) \cdot \Delta t_t
+$$
+
+Each connection receives a unique range of weights, preventing primal degeneracy when multiple connections share a node and have equal marginal cost at the same time step.
```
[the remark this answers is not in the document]
TrentHouliston: :robot: Fixed. Updated the formula to use `p` (priority) instead of `c` (connection index), and clarified that connections with the same priority share a weight range while connections with different priorities get non-overlapping ranges.

docs/modeling/index.md
```diff
@@ -231,6 +242,20 @@ Each element defines its own cost terms based on its role:
...
+Each connection contributes a time-preference term to the secondary objective:
+
+$$
+\text{minimize} \sum_{\text{connections}} \sum_{t=0}^{T-1} w_{c,t} \cdot E_c(t)
+$$
+
+Where $w_{c,t} = c \cdot T + (t + 1)$ assigns monotonically increasing weights per connection $c$ and time step $t$, and $E_c(t)$ is the energy transferred through connection $c$ at time $t$.
+The unique-per-connection weighting prevents primal degeneracy when multiple connections share a node.
+
+The secondary objective does not affect the optimal cost—it only selects among cost-equivalent solutions for deterministic, physically intuitive schedules.
```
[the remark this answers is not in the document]
TrentHouliston: :robot: Fixed. Updated the formula from per-connection index `c` to priority `p`, matching the implementation. Also clarified that priority is auto-computed from endpoint element types.

tests/scenarios/test_benchmark.py
```diff
@@ -0,0 +1,373 @@
...
+        f"  constr={_mean(constraints)}ms"
+        f"  cost={_mean(costs)}ms"
+    )
+
+
+@pytest.mark.scenario
+@pytest.mark.timeout(120)
+@pytest.mark.parametrize(
+    "scenario_path",
+    _scenarios,
```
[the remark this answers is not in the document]
TrentHouliston: :robot: Fixed. Changed from `@pytest.mark.scenario` to `@pytest.mark.benchmark`. The new `benchmark` marker is registered in `pyproject.toml` and is not included in the CI scenario job (`-m scenario`).

tests/scenarios/test_benchmark_sweep.py
```diff
@@ -0,0 +1,239 @@
...
+
+
+@pytest.mark.scenario
+@pytest.mark.timeout(60)
+@pytest.mark.parametrize(
+    "case",
+    SWEEP_CASES,
+    ids=[c.label.replace(" ", "_") for c in SWEEP_CASES],
+)
+def test_sweep_case(
```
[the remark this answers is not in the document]
TrentHouliston: :robot: Fixed. Changed all three test functions from `@pytest.mark.scenario` to `@pytest.mark.benchmark`. The CI scenario job (`-m scenario`) now only collects the 5 snapshot scenario tests.

.github/workflows/ci.yml
```diff
@@ -244,6 +244,22 @@ jobs:
...
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@61cb8a9741eeb8a550a1b8544337180c0fc8476b # v7.2.0
+
+      - name: Install the project
+        run: uv sync --locked --all-extras --dev
+
+      - name: Run scenario tests
+        run: uv run pytest -m scenario -vv
```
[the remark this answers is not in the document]
TrentHouliston: :robot: This is now a non-issue since the benchmark tests have been moved to `@pytest.mark.benchmark`. The CI scenario job (`-m scenario`) only collects the 5 snapshot scenario tests. No changes needed to `ci.yml`.

docs/modeling/index.md
```diff
@@ -75,20 +75,29 @@ The objective minimizes the sum of all element costs.
...
+The primary objective is constrained to its optimal value (within a small epsilon), and the solver minimizes the secondary time-preference objective.
+This breaks ties among cost-equivalent solutions by preferring earlier energy transfers.
+
+**Phase 3: Restore primary duals**
+
+The secondary objective is constrained to its optimal value, and the solver re-minimizes the primary objective.
+This restores dual values (shadow prices) that reflect primary-cost sensitivities at the lexicographically optimal point.
+
+After the initial three-phase solve, subsequent optimizations use a warm-start check.
+If the solver confirms that the current basis is still optimal with no iterations needed and the primary objective value is unchanged, all three phases are skipped.
```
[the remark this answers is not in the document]
TrentHouliston: :robot: Fixed. Updated the docs to match the implementation:
- Phase 2 now correctly says the primary is constrained to its exact optimal value (no epsilon)
- Phase 3 correctly says the secondary is constrained with epsilon slack
- Removed the warm-start skip claim (no such mechanism exists in the current implementation)

.github/workflows/ci.yml
```diff
@@ -229,7 +229,7 @@ jobs:
         run: uv sync --locked --all-extras --dev
 
       - name: Run Tests
-        run: uv run pytest -m "not guide" --cov=custom_components/haeo --cov-branch --cov-report=xml --junitxml=junit.xml -o junit_family=legacy
+        run: uv run pytest -m "not guide and not scenario and not benchmark" --cov=custom_components/haeo --cov-branch --cov-report=xml --junitxml=junit.xml -o junit_family=legacy
```
TrentHouliston, line 232: Benchmarks are skipped by default, no need for the extra complexity.

custom_components/haeo/core/adapters/registry.py
```diff
@@ -14,9 +14,27 @@
...
 from custom_components.haeo.core.model import ModelElementConfig, ModelOutputName
+from custom_components.haeo.core.model.elements import MODEL_ELEMENT_TYPE_CONNECTION
+from custom_components.haeo.core.model.elements.connection import ConnectionElementConfig
 from custom_components.haeo.core.model.output_data import ModelOutputValue, OutputData
 from custom_components.haeo.core.schema.elements import ElementConfigData, ElementType
 
+# Time preference priority for (element_type, direction) pairs.
+# Lower values are preferred by the solver when breaking ties.
+# Direction is relative to the element: "out" = power leaves, "in" = power enters.
+_ENDPOINT_PRIORITY: Final[dict[tuple[ElementType, str], int]] = {
```
TrentHouliston: I generally don't like enumeration. Can you think of another way to do this where the value is stored on the connections themselves? Ideally I would like to do it via properties rather than a hardcoded list. So do something like a lexographical sort on a property which indicates if its production or consumption, a property which indicates if it's your own power or the grid power etc. I'm not 100% fixed on this list order as well. There are some things that are important like solar first before battery discharge and grid export. But other orders are more arbitrary. 

Perhaps the order is "own before external (so use solar/batteries before grid)", "time sensitive before time invariant (I.e. battery can discharge later, solar must go somewhere now)". Maybe there are other properties that help reduce the rest of the order.

custom_components/haeo/core/model/element.py
```diff
@@ -178,40 +178,36 @@ def constraints(self) -> dict[str, highs_cons | list[highs_cons]]:
                     result[name] = cons
         return result
 
-    @cost
-    def cost(self) -> Any:
-        """Return aggregated cost expression from this element.
+    def cost(self) -> list[highs_linear_expression | None] | None:
```
TrentHouliston: Is there a good reason we removed the decorator from this function? That decorator would have prevented this re-running when the downstream changed right?

Why did this function need to change this much in general. It's mostly just changing to deal with the fact that the returned types can have two values. A primary and secondary. Maybe we should just make it explicitly 1/2 elements simplifying some of this logic.

custom_components/haeo/core/model/elements/connection.py
```diff
@@ -36,6 +36,7 @@
 
 CONNECTION_POWER: Final = "connection_power"
 CONNECTION_SEGMENTS: Final = "segments"
+SECONDARY_OBJECTIVE_INDEX: Final = 1
```
TrentHouliston: This might not be needed if we just go to a pure 2 objective style and don't do a generic n long

custom_components/haeo/core/model/elements/connection.py
```diff
@@ -210,22 +214,55 @@ def constraints(self) -> dict[str, highs_cons | list[highs_cons]]:
...
-    def cost(self) -> Any:  # type: ignore[override]
-        """Aggregate costs from all segments."""
-        costs = [sc for seg in self._segments.values() if (sc := seg.cost()) is not None]
+    def cost(self) -> list[highs_linear_expression | None] | None:
+        """Aggregate costs from all segments with multi-objective support.
 
+        Collects costs from all segments and adds tag costs and a time-preference
+        objective to prefer earlier energy transfer when primary costs are equal.
+
+        Time preference is always placed at SECONDARY_OBJECTIVE_INDEX so it
```
TrentHouliston: If we go to a two objective version this would need updating.

custom_components/haeo/core/model/elements/connection.py
```diff
@@ -210,22 +214,55 @@ def constraints(self) -> dict[str, highs_cons | list[highs_cons]]:
         result.update(own_constraints)
         return result
 
-    def cost(self) -> Any:  # type: ignore[override]
-        """Aggregate costs from all segments."""
-        costs = [sc for seg in self._segments.values() if (sc := seg.cost()) is not None]
+    def cost(self) -> list[highs_linear_expression | None] | None:
```
TrentHouliston: Should this be a reactive cost decorator?

custom_components/haeo/core/model/elements/connection.py
```diff
@@ -210,22 +214,55 @@ def constraints(self) -> dict[str, highs_cons | list[highs_cons]]:
...
+                    combined[SECONDARY_OBJECTIVE_INDEX] = Highs.qsum([existing, time_preference])
+
+        return combined or None
 
-        if not costs:
-            return None
-        if len(costs) == 1:
-            return costs[0]
-        return Highs.qsum(costs)
+    def _time_preference_objective(self) -> highs_linear_expression | None:
```
TrentHouliston: This is an entire function for a multiplication. Seems excessive for a one liner.

custom_components/haeo/core/model/elements/segments/segment.py
```diff
@@ -135,24 +135,33 @@ def outputs(self) -> dict[str, OutputData]:
                 result[output_name] = output_data
         return result
 
-    def cost(self) -> Any:
-        """Return aggregated cost expression from this segment."""
-        costs: list[Any] = []
+    def cost(self) -> list[highs_linear_expression | None] | None:
```
TrentHouliston: Once again the reactive decorators are designed to help reduce computation. I'm just checking that this being like this isn't causing issues.

custom_components/haeo/core/model/elements/tests/test_connections.py
```diff
@@ -94,7 +95,10 @@ def _assert_outputs_match(actual: ExpectedOutputFixture, expected: ExpectedOutpu
         assert _is_expected_output(actual)
         assert actual["type"] == expected["type"]
         assert actual["unit"] == expected["unit"]
-        assert actual["values"] == pytest.approx(expected["values"], rel=1e-9, abs=1e-9)
```
TrentHouliston, line 97: This change looks like a ternary on the sensitivity.

Perhaps we should reuse the same logic as the scenarios that are doing dynamic sensitivity.

custom_components/haeo/core/model/network.py
```diff
@@ -18,6 +18,68 @@
 _LOGGER = logging.getLogger(__name__)
 
 
+ObjectiveMode = Literal["lex", "blended", "calibrated"]
+SolverChoice = Literal["simplex", "ipm", "pdlp", "choose"]
+OnOffChoose = Literal["on", "off", "choose"]
```
TrentHouliston, line 23: Are these constants in highs already? I would rather use theirs if they already have a type for this (or maybe define it in the stubs). Except for objective mode, that's ours

custom_components/haeo/core/model/network.py
```diff
@@ -18,6 +18,68 @@
...
+    """Options controlling Network optimization behavior.
+
+    Combines our own multi-objective strategy with a curated set of HiGHS
+    solver options that influence performance on this LP.
+
+    See https://www.gams.com/latest/docs/S_HIGHS.html#HIGHS_OPTIONS for
+    the underlying HiGHS option semantics.
+    """
+
+    # --- Multi-objective strategy (HAEO) ---
```
TrentHouliston: Why does this have haeo in the comment?

custom_components/haeo/core/model/network.py
```diff
@@ -18,6 +18,68 @@
...
+    # blended mode reproduces lex primary variable values within this
+    # fraction of the largest variable magnitude.
+    calibration_tolerance: float = 1e-4
+
+    # --- HiGHS algorithm selection ---
+    # "simplex" is best for warm-starts; "ipm" / "pdlp" do not warm-start.
+    solver: SolverChoice = "simplex"
+    # 1=dual, 4=primal. Primal simplex is ~25-30% faster on cold starts
+    # (benchmarked across all scenarios) and equivalent on warm-starts.
+    simplex_strategy: int = 4
```
TrentHouliston, line 55: Should be an enum. From types from highs directly ideally.

custom_components/haeo/core/model/network.py
```diff
@@ -18,6 +18,68 @@
...
+    blend_weight: float = 1e-3
+    # Relative tolerance on primary decision variable values for
+    # calibrated mode's weight search. The calibration checks that
+    # blended mode reproduces lex primary variable values within this
+    # fraction of the largest variable magnitude.
+    calibration_tolerance: float = 1e-4
+
+    # --- HiGHS algorithm selection ---
+    # "simplex" is best for warm-starts; "ipm" / "pdlp" do not warm-start.
+    solver: SolverChoice = "simplex"
```
TrentHouliston: If the stubs don't have a constant die this. This would be a good spot to define one.

custom_components/haeo/core/model/network.py
```diff
@@ -18,6 +18,68 @@
...
+    # "choose" lets HiGHS skip presolve on warm-starts; "on" forces it.
+    presolve: OnOffChoose = "choose"
+    # "choose" enables parallel only for large problems.
+    parallel: OnOffChoose = "choose"
+    # 0=off, 1=basic, 2=equilibration, 3=forced equilibration.
+    # Scaling off gives a small additional cold-start improvement since
+    # the LP coefficients are already well-conditioned (kW-based units).
+    simplex_scale_strategy: int = 0
+    # Crossover from interior point to a basic solution (ipm/pdlp only).
+    run_crossover: OnOffChoose = "on"
```
TrentHouliston: Having all the options like this feels wrong. Some of them only apply to simplex others to ipm etc. I think it would be better to use discriminated unions to have a specific config object for each of the different flavours.

custom_components/haeo/core/model/network.py
```diff
@@ -18,6 +18,68 @@
...
+
+    def apply(self, solver: Highs) -> None:
+        """Apply HiGHS-tunable options to the given solver."""
+        solver.setOptionValue("solver", self.solver)
+        solver.setOptionValue("simplex_strategy", self.simplex_strategy)
+        solver.setOptionValue("presolve", self.presolve)
+        solver.setOptionValue("parallel", self.parallel)
+        solver.setOptionValue("simplex_scale_strategy", self.simplex_scale_strategy)
+        solver.setOptionValue("run_crossover", self.run_crossover)
```
TrentHouliston: These feel like they should be able to be type checked with overloads etc in the stubs. That would feel safer.

custom_components/haeo/core/model/network.py
```diff
@@ -18,6 +18,68 @@
...
+
+
 @dataclass
```
TrentHouliston: Is this really a data class anymore? It's just feeling like a class to be honest. That we're abusing.

custom_components/haeo/core/model/network.py
```diff
@@ -178,20 +256,258 @@ def optimize(self) -> float:
...
 
-        # Get aggregated cost from network (reactive - only rebuilds if any element cost invalidated)
-        if (total_cost := self.cost()) is not None:
-            h.minimize(total_cost)
-        else:
-            # No cost terms - just run to check feasibility
+        # Get aggregated objectives from network (reactive - only rebuilds if any element cost invalidated)
+        objectives = self.cost()
+
+        _clear_linear_objectives(h)
```
TrentHouliston: Is this required? Surely our updates will already handle this.

custom_components/haeo/core/model/network.py
```diff
@@ -178,20 +256,258 @@ def optimize(self) -> float:
...
+
+        if objectives is None:
+            self._relax_lex_constraint()
```
TrentHouliston: We don't need to do this in blended or calibrated mode typically right?

custom_components/haeo/core/model/network.py
```diff
@@ -178,20 +256,258 @@ def optimize(self) -> float:
...
+        # Pre-compute cost vectors for each objective as dense arrays.
+        # This avoids repeated Python→C FFI overhead from expression-based
+        # setObjective() which zeros all columns then sets non-zero ones.
+        n_vars = h.numVariables
+        all_col_indices = np.arange(n_vars, dtype=np.int32)
+        cost_vectors = _build_cost_vectors(objectives, n_vars)
 
-        msg = f"Optimization failed with status: {h.modelStatusToString(status)}"
-        raise ValueError(msg)
+        h.changeObjectiveSense(ObjSense.kMinimize)
```
TrentHouliston: Aren't we always using the same objective sense? If we are swapping it between primary and secondary that just implies the secondary costs need to be negated right?

custom_components/haeo/core/model/network.py
```diff
@@ -178,20 +256,258 @@ def optimize(self) -> float:
...
+
+        primary = objectives[0]
+        secondary = objectives[1] if len(objectives) > 1 else None
+
+        if primary is None:
+            if secondary is not None:
```
TrentHouliston: There's a lot of logic we're doing for cases where the primary or secondary is none. I think our setup now will never have that. Since any sane network will have connections. That means they will have secondary costs. And a network without primary costs is pointless. I think we treat the primary or secondary or objective none as a "this network isn't ready for optimisation yet" and remove all this weird edge case handling that in practice is meaningless anyway. It's just more code to test that even if it ran wouldn't give a sensible answer since it wasn't asked a sensible question.

custom_components/haeo/core/model/network.py
```diff
@@ -178,20 +256,258 @@ def optimize(self) -> float:
...
+                _set_cost_vector(h, all_col_indices, cost_vectors[1])
+            self._relax_lex_constraint()
```
TrentHouliston: We do this relaxation a lot. But it's not needed in calibrated or blended mode. I think we need to separate out the different modes all in one spot. And relax the lex constraint at the end of the calibration, or at the end of lex etc. wherever it makes sense for each mode.

custom_components/haeo/core/model/network.py
```diff
@@ -178,20 +256,258 @@ def optimize(self) -> float:
...
+            h.run()
+            _ensure_optimal(h)
+            return 0.0
+
+        # Calibrated mode with an existing weight uses blended fast path.
+        use_blended = self.options.mode == "blended" or (
+            self.options.mode == "calibrated" and self._calibrated_weight is not None
+        )
+
+        if use_blended and secondary is not None:
```
TrentHouliston: See heaps of this if secondary and if primary. I think we just assume both always exist.

custom_components/haeo/core/model/network.py
```diff
@@ -178,20 +256,258 @@ def optimize(self) -> float:
...
+            if self.options.mode == "calibrated" and self._calibrated_weight is not None:
+                return self._solve_blended(h, all_col_indices, cost_vectors, self._calibrated_weight)
+            return self._solve_blended(h, all_col_indices, cost_vectors, self.options.blend_weight)
+
+        # --- Lexicographic solve (also used for calibrated-mode first call) ---
```
TrentHouliston: This seems like a bit of a mess. The blended and calibrated solves are in functions but the lex isn't? This code can be organised much better.

custom_components/haeo/core/model/network.py
```diff
@@ -178,20 +256,258 @@ def optimize(self) -> float:
...
+        changes shifting the boundary.
+        """
+        h = self._solver
+
+        # Primary variable indices: nonzero coefficient in primary cost vector.
+        pri_mask = cost_vectors[0] != 0
+        lex_pri = lex_values[pri_mask]
+
+        # If no variables have primary cost, any weight is safe.
+        if lex_pri.size == 0:
```
TrentHouliston, line 399: As said earlier. Let's put a guard right up the top that if we don't have enough elements for a primary and secondary just return that the optimisation can't happen yet. That will remove so many if statements from the rest of this class

custom_components/haeo/core/model/network.py
```diff
@@ -178,20 +256,258 @@ def optimize(self) -> float:
...
+                return False
+            bl_vals = np.asarray(h.allVariableValues())
+            max_diff = float(np.max(np.abs(bl_vals[pri_mask] - lex_pri)))
+            return max_diff <= abs_tol
+
+        lo, hi = _CAL_LOG_LO, _CAL_LOG_HI
+        half_budget = _CAL_MAX_STEPS // 2
+
+        # --- Find upper edge: highest weight where primary vars match ---
+        if _primary_vars_match(hi):
```
TrentHouliston, line 421: Don't there already exist functions to do binary search style logic in Python rather than implementing it ourself? At the very least I feel like the binary search part should be separated out. It's just a generic algorithm.

custom_components/haeo/core/model/network.py
```diff
@@ -206,3 +522,68 @@ def constraints(self) -> dict[str, dict[str, highs_cons | list[highs_cons]]]:
             if element_constraints := element.constraints():
                 result[element_name] = element_constraints
         return result
+
+
+def _clear_linear_objectives(solver: Highs) -> None:
```
TrentHouliston: A function for a one liner seems excessive.

custom_components/haeo/core/model/network.py
```diff
@@ -206,3 +522,68 @@ def constraints(self) -> dict[str, dict[str, highs_cons | list[highs_cons]]]:
...
+    """Clear any multiobjective state."""
+    solver.clearLinearObjectives()
+
+
+def _build_cost_vectors(
```
TrentHouliston, line 562: Doesn't highs already do this with its affine expression things?

custom_components/haeo/core/model/network.py
```diff
@@ -206,3 +522,68 @@ def constraints(self) -> dict[str, dict[str, highs_cons | list[highs_cons]]]:
...
+    return vectors
+
+
+def _set_cost_vector(
+    solver: Highs,
+    col_indices: NDArray[np.int32],
+    costs: NDArray[np.float64],
+) -> None:
+    """Set the full objective cost vector in a single C call."""
+    solver.changeColsCost(len(col_indices), col_indices, costs)
```
TrentHouliston, line 593: Is there a potential bug here that you're not setting everything? Could some old costs hang around?

custom_components/haeo/core/model/network.py
```diff
@@ -206,3 +522,68 @@ def constraints(self) -> dict[str, dict[str, highs_cons | list[highs_cons]]]:
...
+
+    The epsilon must be large enough to guarantee the constraint has slack
+    (making the dual structurally zero) but small enough not to
+    meaningfully shift the optimal vertex. A relative tolerance scaled by
+    the objective magnitude achieves this.
+    """
+    return max(1e-6, abs(value) * 1e-6)
+
+
+# Calibration search bounds in log10 space.  The secondary objective can
```
TrentHouliston: Constants at the end of a file are weird. Also wouldn't these be class member variables?

custom_components/haeo/core/model/element.py
```diff
@@ -180,38 +180,35 @@ def constraints(self) -> dict[str, highs_cons | list[highs_cons]]:
 
     @cost
     def cost(self) -> Any:
```
TrentHouliston: We shouldn't need any on the costs right? They should be either a highsexpression or tuple of highs expression or none, or none. 

HighsExpression | (HighsExpression | None,) | None

custom_components/haeo/core/model/element.py
```diff
@@ -180,38 +180,35 @@ def constraints(self) -> dict[str, highs_cons | list[highs_cons]]:
...
+        recomputes when underlying @cost method dependencies change.
 
         Returns:
-            Single aggregated cost expression (highs_linear_expression) or None if no costs
+            Single cost expression or None if no costs are defined
 
         """
-        # Get this method's name from the decorator to avoid hardcoding
-        this_method_name = type(self).cost._name  # type: ignore[attr-defined]  # noqa: SLF001 (intentional access to decorator's name)
+        this_method_name = type(self).cost._name  # type: ignore[attr-defined]  # noqa: SLF001
```
TrentHouliston, line 191: Any noqa must have an explanatory comment describing why no other alternative was possible.

custom_components/haeo/core/model/element.py
```diff
@@ -180,38 +180,35 @@ def constraints(self) -> dict[str, highs_cons | list[highs_cons]]:
...
             method = getattr(self, name)
             if (cost_value := method()) is not None:
                 costs.append(cost_value)
 
-        # Aggregate costs into a single expression
         if not costs:
             return None
-        return reduce(operator.add, costs)
+        if len(costs) == 1:
+            return costs[0]
```
TrentHouliston, line 208: I think we should change our types so it's either a single instance, or a tuple of two. So no single length enforced at type time.

custom_components/haeo/core/model/elements/connection.py
```diff
@@ -101,6 +104,16 @@ def segments(self) -> OrderedDict[str, Segment]:
...
         return self._segments
 
+    @property
+    def sort_key(self) -> tuple[int, str, str, str]:
+        """Deterministic sort key for time-preference ordering.
+
+        Sorts by adapter-assigned priority first (from endpoint element types
+        and directions), then by source, target, and name for deterministic
+        tiebreaking between connections at the same priority level.
+        """
```
TrentHouliston, line 117: This isn't the sort order I was suggesting. Take a look at my original review comment describing the order. It involved external vs internal etc. some of which may require new attributes.

custom_components/haeo/core/model/elements/connection.py
```diff
@@ -210,22 +223,35 @@ def constraints(self) -> dict[str, highs_cons | list[highs_cons]]:
         result.update(own_constraints)
         return result
 
-    def cost(self) -> Any:  # type: ignore[override]
-        """Aggregate costs from all segments."""
-        costs = [sc for seg in self._segments.values() if (sc := seg.cost()) is not None]
+    def cost(self) -> tuple[Any, Any] | None:  # type: ignore[override]
```
TrentHouliston: Should this have a cost decorator?

custom_components/haeo/core/model/elements/segments/segment.py
```diff
@@ -135,19 +141,22 @@ def outputs(self) -> dict[str, OutputData]:
                 result[output_name] = output_data
         return result
 
+    @cost
     def cost(self) -> Any:
```
TrentHouliston: I think in general we need to improve the typing on costs. There is no good reason that we should not be typing here.

custom_components/haeo/core/model/elements/tests/test_elements.py
```diff
@@ -61,8 +61,14 @@ def mock_connection_power() -> HighspyArray:
...
-        if (element_cost := element.cost()) is not None:
-            element_costs.append(element_cost)
+        element_cost = element.cost()
+        if element_cost is not None:
+            if isinstance(element_cost, tuple):
+                pri, _sec = element_cost
+                if pri is not None:
+                    element_costs.append(pri)
+            else:
+                element_costs.append(element_cost)
```
TrentHouliston: None is not a tuple etc. this can be simplified a lot.

stubs/highspy/__init__.pyi
```diff
@@ -100,7 +124,7 @@ class Highs:
     def numConstrs(self) -> int: ...
     def __init__(self) -> None: ...
     def setOptionValue(self, option: str, value: bool | int | float | str) -> None: ...
-    def getOptionValue(self, option: str) -> bool | int | float | str: ...
+    def getOptionValue(self, option: str) -> tuple[HighsStatus, bool | int | float | str]: ...
```
TrentHouliston, line 143: I think this can be done better with overloads and literals instead of str for the option. That would mean the typing is consistent. Same for setting options.

custom_components/haeo/core/model/reactive/tests/test_warm_start.py
```diff
@@ -255,7 +265,11 @@ def test_connection_update_max_power_target_source() -> None:
                 "power_limit": {
                     "segment_type": "power_limit",
                     "max_power": 5.0,
-                }
+                },
+                "pricing": {
+                    "segment_type": "pricing",
+                    "price": 0.01,
```
TrentHouliston: Best to leave off the trailing comma in this (and the power limit above and similar cases) that way ruff will put it on one line

custom_components/haeo/core/model/network.py
```diff
@@ -23,61 +23,79 @@
...
-@dataclass(frozen=True)
-class SolveOptions:
-    """Options controlling Network optimization behavior.
+# Calibration search bounds in log10 space.  The secondary objective can
+# be many orders of magnitude larger than the primary, so we need a wide
+# range.  1e-12 is effectively zero influence; 1e-1 would dominate.
+_CAL_LOG_LO = -12.0
+_CAL_LOG_HI = -1.0
+_CAL_MAX_STEPS = 40  # total bisection budget split across upper/lower searches
+_CAL_CONVERGENCE = 0.01  # stop bisection when interval < this (log10 decades)
```
TrentHouliston: When making constants like this you should type them Final, so we don't accidentally change them later. I noticed these were duplicated at one point in the commit chain

custom_components/haeo/core/model/network.py
```diff
@@ -115,6 +133,9 @@ def __post_init__(self) -> None:
         # Apply tunable solver options
         self.options.apply(self._solver)
 
+        # Always minimize — set once, never changed
+        self._solver.changeObjectiveSense(ObjSense.kMinimize)
+
```
TrentHouliston: Isn't minimize the default?

custom_components/haeo/core/model/network.py
```diff
@@ -201,128 +222,113 @@ def add(self, element_config: ModelElementConfig) -> Element[Any]:
...
         if objectives is None:
-            self._relax_lex_constraint()
-            h.run()
-            return _ensure_optimal(h)
+            msg = "Network has no cost objectives — add connections with pricing segments"
+            raise ValueError(msg)
+
+        primary, secondary = objectives
+        if primary is None:
+            msg = "Network has no primary cost — add pricing segments to connections"
```
TrentHouliston: It's not just pricing segments on connections, Pricing can also come from nodes too.

tests/scenarios/test_benchmark.py
```diff
@@ -0,0 +1,238 @@
...
+    ids=[s.name for s in _scenarios],
+    indirect=["scenario_path"],
+)
+@pytest.mark.parametrize("options", _MODES, ids=[m.mode for m in _MODES])
+class TestBenchmark:
+    """Optimization benchmarks for each scenario and solve mode."""
```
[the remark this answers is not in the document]
TrentHouliston: Indeed this still needs fixing

PR #364, opened 2026-04-12
refactor: unidirectional connections and functional segment composition

TrentHouliston:
## Summary

Two architectural changes to the model layer:

1. **Connections are now unidirectional.** Each `Connection` represents a single direction of power flow (source → target). Bidirectional paths (grid import/export, battery charge/discharge, inverter DC↔AC) are modeled as two separate `Connection` elements with a naming convention (e.g., `grid:import` / `grid:export`, `battery:discharge` / `battery:charge`).

2. **Segments are functional transforms, not variable owners.** `Connection` creates LP variables for the power flow, then chains them through segments via `power_in` / `power_out`. Segments add constraints and costs as side effects, returning (possibly transformed) power expressions. No segment creates its own power-flow variables.

### Before / after

| | Before | After |
|---|---|---|
| **Connection** | Bidirectional — created variables for both directions, linking constraints between segments | Unidirectional — single variable array, chained through segments |
| **Segments** | Each segment created its own per-direction variables (`*_source_target`, `*_target_source`) | Single-direction params (`max_power`, `price`, `efficiency`), receive `power_in` at init |
| **Bidirectional flows** | One connection with bidirectional segments | Two connections (forward + reverse) |
| **Adapters** | Single connection element per device | Forward/reverse (or import/export, charge/discharge) connection pairs |

### Segment types

| Segment | Classification | Behavior |
|---|---|---|
| `PassthroughSegment` | Identity | Returns input unchanged |
| `PricingSegment` | Identity + cost | Adds cost expression using `price × power_in × period` |
| `PowerLimitSegment` | Identity + constraint | Adds upper bound constraint on `power_in` |
| `EfficiencySegment` | Transform | Returns `power_in × efficiency` |
| `SocPricingSegment` | Auxiliary | Adds slack variables + penalty costs for SOC thresholds |

### Variable reduction

For a 3-period connection with power_limit + pricing segments:
- **Before:** 12 variables (2 directions × 2 segments × 3 periods) + 6 linking constraints
- **After:** 3 variables (1 direction × 3 periods), zero linking constraints

### Model layer (`core/model/`)
- `Connection` is now unidirectional with a single `power_in` variable array chained through segments
- All segment params simplified from per-direction (`*_source_target` / `*_target_source`) to single values (`max_power`, `price`, `efficiency`)
- Removed time-slice constraint from `PowerLimitSegment`
- Removed `mirror_segment_order` configuration
- Connection output names simplified: `connection_power` replaces `connection_power_source_target` / `connection_power_target_source`

### Adapter layer (`core/adapters/`)
- **Battery:** Split into `discharge` (battery → network) and `charge` (network → battery) connections; SOC pricing only on discharge
- **Grid:** Split into `import` (grid → network) and `export` (network → grid) connections
- **Inverter:** Split into `connection` (DC → AC) and `reverse` (AC → DC) connections
- **Load:** Single connection (network → load), direction reversed from previous source orientation
- **Solar:** Single connection (solar → network), pricing segment conditionally included
- **Connection (generic):** Split into `forward` and `reverse` connections
- Shadow price extraction updated to use segment-level `power_limit` output key

### Other changes
- Removed Nord Pool price extractor (separate cleanup)
- Simplified guide runner tooling (removed hashing, simplified fence processing)
- Updated scenario test snapshots and visualizations for new connection topology
- Updated documentation for unidirectional model

## Test impact

75 files changed, ~4000 insertions / ~4500 deletions (net reduction). All existing tests adapted to the new architecture.


custom_components/haeo/core/adapters/elements/tests/test_battery.py
```diff
@@ -381,14 +381,14 @@ def test_model_elements_omits_efficiency_when_missing() -> None:
...
+    connection = _get_connection(elements, "test_battery:discharge")
     segments = connection.get("segments")
     assert segments is not None
     efficiency_segment = segments.get("efficiency")
     assert efficiency_segment is not None
     assert is_efficiency_spec(efficiency_segment)
-    assert efficiency_segment.get("efficiency_source_target") is None
-    assert efficiency_segment.get("efficiency_target_source") is None
+    assert efficiency_segment.get("efficiency") is None
+    assert efficiency_segment.get("efficiency") is None
```
TrentHouliston: Is this a duplicated line?

custom_components/haeo/core/adapters/elements/battery.py
```diff
@@ -181,33 +181,36 @@ def model_elements(self, config: BatteryConfigData) -> list[ModelElementConfig]:
...
+        # Discharge: battery -> network
         elements.append(
             {
                 "element_type": MODEL_ELEMENT_TYPE_CONNECTION,
-                "name": f"{name}:connection",
+                "name": f"{name}:discharge",
                 "source": name,
                 "target": extract_connection_target(config[CONF_CONNECTION]),
-                "segments": segments,
+                "segments": discharge_segments,
```
TrentHouliston, line 199: Just inline these, the charging segments below are inlined. Be consistent, also pick inlining all other things being equal for this kind of thing

custom_components/haeo/core/adapters/elements/inverter.py
```diff
@@ -67,45 +62,42 @@ class InverterAdapter:
...
-        max_power_source_target = power_limits.get(CONF_MAX_POWER_SOURCE_TARGET)
-        max_power_target_source = power_limits.get(CONF_MAX_POWER_TARGET_SOURCE)
-        if max_power_source_target is None or max_power_target_source is None:
-            msg = "Inverter power limits missing - config flow validation failed"
-            raise RuntimeError(msg)
-
+        """Return model element parameters for Inverter configuration."""
+        max_power_st = config[SECTION_POWER_LIMITS].get(CONF_MAX_POWER_SOURCE_TARGET)
+        max_power_ts = config[SECTION_POWER_LIMITS].get(CONF_MAX_POWER_TARGET_SOURCE)
+        inv_name = config["name"]
```
TrentHouliston: These are all used in one spot (or close enough) we can just linline it all

custom_components/haeo/core/adapters/elements/connection.py
TrentHouliston, line 1: Just make this connection and the corresponding flow/element/etc be single directional. Make a migration that makes the other direction. This is an advanced element type that typically won't be made directly anyway.

custom_components/haeo/core/adapters/elements/inverter.py
```diff
@@ -67,45 +62,42 @@ class InverterAdapter:
...
-                    "power_limit": {
-                        "segment_type": "power_limit",
-                        "max_power_source_target": max_power_source_target,
-                        "max_power_target_source": max_power_target_source,
+                    "power_limit": {"segment_type": "power_limit", "max_power": max_power_st},
+                },
+            },
+            {
+                "element_type": MODEL_ELEMENT_TYPE_CONNECTION,
+                "name": f"{inv_name}:reverse",
```
TrentHouliston: Maybe instead of :connection and :reverse we should name based on the direction (ac_to_dc/dc_to_ac, or maybe invert/rectify or something like that). Just to fit in with charge/discharge import/export etc

custom_components/haeo/core/adapters/elements/load.py
```diff
@@ -57,31 +56,32 @@ def model_elements(self, config: LoadConfigData) -> list[ModelElementConfig]:
...
+                "source": target_name,
+                "target": load_name,
                 "segments": {
                     "power_limit": {
                         "segment_type": "power_limit",
-                        "max_power_source_target": 0.0,
-                        "max_power_target_source": config[SECTION_FORECAST][CONF_FORECAST],
-                        "fixed": not config[SECTION_CURTAILMENT].get(CONF_CURTAILMENT, False),
+                        "max_power": forecast,
+                        "fixed": fixed,
```
TrentHouliston: When values are only used once like this, just inline them like they were before. Having a line above that exists just to extract a value that's used once just breaks up the flow

custom_components/haeo/core/adapters/elements/load.py
```diff
@@ -96,16 +96,16 @@ def outputs(
         """Map model outputs to load-specific output names."""
         connection = model_outputs[f"{name}:connection"]
 
-        power_target_source = expect_output_data(connection[CONNECTION_POWER_TARGET_SOURCE])
+        power = expect_output_data(connection[CONNECTION_POWER])
```
TrentHouliston, line 93: Yeah, just go through most of these and find the ones that are single use variables declared above a structured output, and just use the value directly

custom_components/haeo/translations/en.json
```diff
@@ -475,12 +475,8 @@
       "battery_soc_max": { "name": "SOC max shadow price" },
       "battery_soc_min": { "name": "SOC min shadow price" },
       "battery_state_of_charge": { "name": "State of charge" },
+      "connection_power": { "name": "{source} to {target} power" },
       "connection_power_active": { "name": "{source} to {target} active power" },
-      "connection_power_source_target": { "name": "{source} to {target} power" },
-      "connection_power_target_source": { "name": "{target} to {source} power" },
-      "connection_shadow_power_max_source_target": { "name": "Max {source} to {target} power shadow price" },
```
TrentHouliston, line 481: I understand the removal of most of these, but isn't there still a shadow price for the power limit on a connection?

custom_components/haeo/core/model/tests/test_data/segments/pricing.py
```diff
@@ -15,15 +15,14 @@
...
+            "price": np.array([0.1, 0.1]),
         },
         "periods": np.array([1.0, 0.5]),
         "inputs": {
             "power_in_st": [10.0, 10.0],
-            "power_in_ts": [5.0, 5.0],
             "minimize_cost": True,
         },
-        "expected_outputs": {"objective_value": 3.0},
+        # cost = 10 * 0.1 * 1.0 + 10 * 0.1 * 0.5 = 1.0 + 0.5 = 1.5
```
TrentHouliston: Don't put math in comments, it's brittle

custom_components/haeo/core/model/elements/segments/tests/test_segments.py
```diff
@@ -182,7 +199,15 @@ def _solve_segment_scenario(case: SegmentScenario) -> dict[str, ExpectedValue]:
         if key == "objective_value":
             outputs[key] = float(h.getObjectiveValue())
             continue
-        flow = getattr(seg, key)
+        # Map directional keys to the correct single-direction segment
+        if key.endswith("_st"):
```
TrentHouliston: We shouldn't need _st or _ts anymore. These seem like redundant fallbacks that should be removed. We should be cleanly mapping from end to end without key manipulation

custom_components/haeo/core/adapters/elements/solar.py
```diff
@@ -93,16 +87,16 @@ def outputs(
...
-            SOLAR_POWER: replace(power_source_target, type=OutputType.POWER),
+            SOLAR_POWER: replace(power, type=OutputType.POWER),
         }
 
         # Shadow price from power_limit segment (if present)
         if (
             isinstance(segments_output := connection.get(CONNECTION_SEGMENTS), Mapping)
             and isinstance(power_limit_outputs := segments_output.get("power_limit"), Mapping)
-            and (shadow := expect_output_data(power_limit_outputs.get(POWER_LIMIT_SOURCE_TARGET))) is not None
+            and (shadow := expect_output_data(power_limit_outputs.get("power_limit"))) is not None
```
TrentHouliston, line 96: Should some of these string literals you've added back in be done using the constants?

custom_components/haeo/core/adapters/elements/solar.py
```diff
@@ -56,31 +50,31 @@ class SolarAdapter:
...
+        solar_name = config["name"]
+        target_name = extract_connection_target(config[CONF_CONNECTION])
+        price = config[SECTION_PRICING].get(CONF_PRICE_SOURCE_TARGET)
+        segments: dict[str, Any] = {
+            "power_limit": {
+                "segment_type": "power_limit",
+                "max_power": config[SECTION_FORECAST][CONF_FORECAST],
+                "fixed": not config[SECTION_CURTAILMENT].get(CONF_CURTAILMENT, True),
+            },
+        }
```
TrentHouliston: Just inline all this if you can, I'd prefer a single inline dictionary as much as possible

PR #370, opened 2026-04-15
Add tag infrastructure to model layer

TrentHouliston:
Add per-source power tagging (VLAN) infrastructure to the model layer, enabling per-tag power flow decomposition and per-source cost differentiation.

### Element base (`element.py`)
- Add `outbound_tags` and `inbound_tags` on `NetworkElement` for tag-based power routing
- Add `element_power_produced()` / `element_power_consumed()` protocol — returns `None` when not applicable, HighspyArray when active
- Add `element_power_balance` constraint: per-tag power balance with outbound_tags/inbound_tags enforcement
- Production decomposed into per-tag `produced_by_tag` LP variables (like `consumed_by_tag`)
- Per-connection blocking: blocked tags force each connection's per-tag flow to zero individually (prevents pass-through)
- Empty tag set guards: if no outbound/inbound overlap exists, production/consumption forced to zero
- Add `connection_power_for_tag()` and `connection_tags()` queries
- Cost aggregation simplified with `reduce(operator.add, costs)` — no branching

### Node (`node.py`)
- Separate `_produced` / `_consumed` LP variables encoding source/sink behavior
- Non-source nodes return `None` for production, non-sink nodes return `None` for consumption
- Source+sink skips total balance (unconstrained)

### Battery (`battery.py`)
- `element_power_produced()` returns discharge power, `element_power_consumed()` returns charge power
- Power balance uses produced/consumed form

### Connection (`connection.py`)
- Per-tag LP variable decomposition: tag flows sum to total flow
- Per-tag segment transforms (efficiency applied proportionally)
- `tag_prices` for per-source pricing on connections (pricing segment parameter)
- `power_into_source_for_tag()` / `power_into_target_for_tag()` queries

### Tests
- `test_policy_scenario.py`: Full tagged power flow scenario (Grid/Solar/Battery/Load with 3 tags and per-tag pricing)
- `test_element.py`: TDD tests for production duplication, empty inbound/outbound guards, per-connection tag blocking, passthrough elements, untagged connections
- `test_node_model.py`: Unconstrained node adapter output case
- `test_segments.py`: Tag transfer cost with no matching tags

### Documentation
- `tagged-power.md`: Tag parameter reference and design overview
- Test instructions updated with coverage decision tree and bug-fix TDD guidance


custom_components/haeo/core/model/tests/test_data/node.py
```diff
@@ -22,7 +22,7 @@
         "data": {
             "name": "hub_node",
             "periods": np.array([1.0] * 3),
-            "is_source": False,
+            "is_source": True,
```
TrentHouliston, line 25: Why did this change?

custom_components/haeo/core/model/elements/node.py
```diff
@@ -52,42 +53,33 @@ def __init__(
...
-            return list(conn_power >= 0)
-        # Can both produce and consume power so there are no bounds
-        return None
+        n = self.n_periods
+        self._produced: HighspyArray | None = None
+        self._consumed: HighspyArray | None = None
+        if is_source:
+            self._produced = solver.addVariables(n, lb=0, name_prefix=f"{name}_prod_", out_array=True)
+        if is_sink:
+            self._consumed = solver.addVariables(n, lb=0, name_prefix=f"{name}_cons_", out_array=True)
```
TrentHouliston: May as well just make a highspy expression array here directly and just set it to zeros or the variables rather than checking each time in the element_power_produced/element_power_consumed functions

custom_components/haeo/core/model/element.py
```diff
@@ -119,6 +128,26 @@ def n_periods(self) -> int:
...
+
+    @property
+    def access_list(self) -> set[int] | None:
+        """Return the set of tags this element can consume, or None for all."""
+        return self._access_list
+
+    @access_list.setter
+    def access_list(self, value: set[int] | list[int] | None) -> None:
+        """Set the access list."""
+        self._access_list = set(value) if value else None
```
TrentHouliston: Is there any advantage to setters/getters for this over just storing the variables? Currently it looks like extra boilerplate to me. I would suggest deleting the setter/getters and just directly using the variables.

custom_components/haeo/core/model/network.py
```diff
@@ -195,7 +195,7 @@ def optimize(self) -> float:
 
         # Check optimization status
         status = h.getModelStatus()
-        if status == HighsModelStatus.kOptimal:
+        if status in (HighsModelStatus.kOptimal, HighsModelStatus.kModelEmpty):
```
TrentHouliston: I find this suspicious. Having an empty model feels like an error, not success

docs/modeling/model-layer/connections/connection.md
```diff
@@ -100,6 +102,54 @@ Connection exposes power flow and segment outputs:
...
+
+This enables per-source pricing (e.g., charging a premium for battery-sourced power exported to the grid).
+
+### Per-tag power balance interface
+
+Connections provide per-tag versions of the power balance interface used by elements:
+
+$$P_{\text{into\_source},k}(t) = -P_{\text{in},k}(t)$$
+
+$$P_{\text{into\_target},k}(t) = P_{\text{out},k}(t)$$
```
TrentHouliston, line 148: connection.md documentation talks a lot about specific segments. Connection should be agnostic to the segments it holds, therefore the docs should describe how connection works and interacts with segments, not how the segments work.

docs/modeling/model-layer/connections/connection.md
```diff
@@ -100,6 +102,54 @@ Connection exposes power flow and segment outputs:
 The `segments` output groups segment outputs using the segment names provided in the configuration.
 Adapters use this map to surface segment-specific shadow prices.
 
+## Tag decomposition
+
+When `tags` are provided, the connection decomposes its power flow into per-tag LP variables.
+This enables [tagged power](../../tagged-power.md) tracking and per-source cost differentiation.
```
TrentHouliston: I see things like

> When `tags` are provided, the connection decomposes its power flow into per-tag LP variables.

scattered around throughout the code. I think we should just commit to taggd power. If we wanted to make an untagged version, we just provde a single tag for everything. That way the logic is all consistent.

custom_components/haeo/core/model/element.py
```diff
@@ -36,6 +39,8 @@ def __init__(
         *,
         solver: Highs,
         output_names: frozenset[OutputNameT],
+        source_tag: int | None = None,
+        access_list: list[int] | None = None,
```
TrentHouliston: The source_tag and access_list is really just "what vlans can output/input reach". We can probably make them consistent and just have an inbound_tags and outbound_tags. Then both can be arrays/sets and it'll be consistent that the different in/outs are just able to put their power on specific vlans. We can scope it differently when we get to the policy layer, but for the model layer it's probably easy enough to just make it consistent.

custom_components/haeo/core/model/elements/segments/segment.py
```diff
@@ -42,6 +42,7 @@ def __init__(
         source_element: Element[Any],
         target_element: Element[Any],
         power_in: HighspyArray,
+        tag_flows_in: dict[int, HighspyArray] | None = None,
```
TrentHouliston: Segments shouldn't need a separate power_in and tag_flows_in. Given that they should be passing along the power dictionaries, they should only be receiving those power dictionaries. If they need the sum they can do it themselves.

custom_components/haeo/core/model/elements/battery.py
```diff
@@ -53,7 +53,7 @@
 )
 
 # Battery power constraints (subset of outputs that relate to power balance)
-BATTERY_POWER_CONSTRAINTS: Final[frozenset[BatteryConstraintName]] = frozenset((BATTERY_POWER_BALANCE,))
```
TrentHouliston, line 56: This is a regression in the types here

custom_components/haeo/core/model/elements/segments/__init__.py
```diff
@@ -87,6 +87,7 @@ def create_segment(
     source_element: Element[Any],
     target_element: Element[Any],
     power_in: HighspyArray,
+    tag_flows_in: dict[int, HighspyArray] | None = None,
```
TrentHouliston: As specified elsewhere, I don't think these should be split

custom_components/haeo/core/model/elements/segments/efficiency.py
```diff
@@ -64,5 +66,12 @@ def power_out(self) -> HighspyArray:
             return self._power_in
         return self._power_in * self.efficiency
 
+    @property
+    def tag_flows_out(self) -> dict[int, HighspyArray]:
```
TrentHouliston: I think we should only have a flows_in and flows_out. We always need to traverse the tags path so it may as well be the only way. Passing out the totals doens't have much use. We can sum the tags whenever we need to.

custom_components/haeo/core/model/elements/segments/passthrough.py
```diff
@@ -33,6 +33,7 @@ def __init__(
         source_element: Element[Any],
         target_element: Element[Any],
         power_in: HighspyArray,
+        tag_flows_in: dict[int, HighspyArray] | None = None,
```
TrentHouliston: Same for this and all the others I've mentioned. only have power_in which is the tagged flows. No need for the "summed" version as we want to operate on the tagged values always anyway.

custom_components/haeo/core/model/elements/segments/pricing.py
```diff
@@ -55,8 +57,10 @@ def __init__(
             source_element=source_element,
             target_element=target_element,
             power_in=power_in,
+            tag_flows_in=tag_flows_in,
         )
         self.price = broadcast_to_sequence(spec.get("price"), self._n_periods)
+        self._tag_costs: list[dict[str, Any]] = spec.get("tag_costs") or []
```
TrentHouliston: tag_prices if anything. We specifically use prices for how much per unit something is valued at. Costs are in $ and they are the actual paid amount.

custom_components/haeo/core/model/elements/segments/pricing.py
```diff
@@ -65,5 +69,21 @@ def transfer_cost(self) -> highs_linear_expression | None:
...
+        costs: list[highs_linear_expression] = []
+        for tc in self._tag_costs:
+            tag = tc["tag"]
+            price = broadcast_to_sequence(tc.get("price"), self._n_periods)
+            if price is not None and tag in self._tag_flows_in:
+                tag_flow = self._tag_flows_in[tag]
+                costs.append(Highs.qsum(tag_flow * price * self.periods))
+        if not costs:
+            return None
+        return Highs.qsum(costs) if len(costs) > 1 else costs[0]
```
TrentHouliston: This feels like it can be a list comprehension of some sort. This seems very complex for what it is.

custom_components/haeo/core/model/elements/connection.py
```diff
@@ -5,6 +5,10 @@
 
 Connection creates LP variables for the power flow, then chains them through
 segments. Each segment may add constraints, costs, or transform the expression.
+
+When tags are provided, the flow is decomposed into per-tag LP variables that
+sum to the total power. Segments are tag-aware and handle per-tag transforms
+and costs internally.
```
TrentHouliston: Once again, this comment "when tags are provided" implies there is a "non tag" path. If we go with tags for everything which will make the handling simpler, there are always tags, even if it's just a single "default" tag.

custom_components/haeo/core/model/elements/connection.py
```diff
@@ -86,6 +92,11 @@ def __init__(
         self._power_in: HighspyArray | None = None
         self._power_out: HighspyArray | None = None
```
TrentHouliston: These should just be the tagged dicts. There should only be the tagged versions (named power_in and power_out).

The totals are not useful for us if we are always doing tagged power

custom_components/haeo/core/model/elements/connection.py
```diff
@@ -109,12 +120,28 @@ def set_endpoints(self, source_element: Element[Any], target_element: Element[An
...
+                    self.n_periods, lb=0, name_prefix=f"{self.name}_t{tag}_", out_array=True,
+                )
+
+            # Total power_in is the sum of all tag flows
+            self._power_in = self._solver.addVariables(
+                self.n_periods, lb=0, name_prefix=f"{self.name}_", out_array=True,
+            )
+            self._solver.addConstrs(
+                self._power_in == sum(tag_flows.values())  # type: ignore[arg-type]
+            )
```
TrentHouliston: Not needed if we go tags only

custom_components/haeo/core/model/elements/connection.py
```diff
@@ -146,6 +177,18 @@ def power_out(self) -> HighspyArray:
...
+        """Return the set of tags on this connection."""
+        return self._tags
+
+    def power_into_source_for_tag(self, tag: int) -> HighspyArray:
+        """Power flowing into the source node for a specific tag."""
+        return -self._tag_power_in[tag]
+
+    def power_into_target_for_tag(self, tag: int) -> HighspyArray:
+        """Power flowing into the target node for a specific tag."""
+        return self._tag_power_out[tag]
```
TrentHouliston: Once again, this is dealing with tagged vs untagged variants. We only have tagged variants

custom_components/haeo/core/model/elements/node.py
```diff
@@ -3,21 +3,18 @@
...
 type NodeElementTypeName = Literal["node"]
 ELEMENT_TYPE: Final[NodeElementTypeName] = "node"
 
-type NodeConstraintName = Literal["node_power_balance"]
+type NodeOutputName = Literal["element_power_balance"]
 
-type NodeOutputName = NodeConstraintName
-
-NODE_POWER_BALANCE: Final[NodeOutputName] = "node_power_balance"
+NODE_POWER_BALANCE: Final[str] = ELEMENT_POWER_BALANCE
```
TrentHouliston: Type regression making this an str rather than the specific type. I've seen a few cases in the diff where you've regressed the types to less specific types.

custom_components/haeo/core/model/elements/node.py
```diff
@@ -29,19 +26,23 @@ class NodeElementConfig(TypedDict):
...
+    access_list: NotRequired[list[int] | None]
 
 
 class Node(Element[NodeOutputName]):
     """Node entity for electrical system modeling.
 
-    Node acts as an infinite source and/or sink. Power limits and pricing are configured
-    on the Connection to/from the node.
+    Node acts as an infinite source and/or sink. Power limits and pricing are
+    configured on the Connection to/from the node.
```
TrentHouliston, line 37: This comment is irrelevant to this class. You shouldn't need to discuss what you can do with other classes in this one. This should be documenting this class.

custom_components/haeo/core/model/tests/test_element.py
```diff
@@ -84,16 +84,16 @@ def test_constraints_populates_constraint_state(solver: Highs) -> None:
...
 
     # Call constraints to trigger decorator lifecycle
     constraints_dict = element.constraints()
 
     # After calling, constraint state should exist
-    # Node has node_power_balance (returns constraints when is_source=True, is_sink=False)
-    assert "node_power_balance" in constraints_dict
-    assert constraints_dict["node_power_balance"] is not None
+    # Node has element_power_balance from Element base (creates constraints when not both source+sink)
+    assert "element_power_balance" in constraints_dict
```
TrentHouliston: We should be using constants rather than literals. "element_power_balance" is a constant somewhere

custom_components/haeo/core/model/element.py
```diff
@@ -129,6 +158,28 @@ def register_connection(self, connection: Any, end: Literal["source", "target"])
...
+
+        Positive values represent power absorbed from the network.
+        Consumption is distributed across allowed tags (per ``access_list``).
+        Override in subclasses.  Default returns 0.
+        """
+        return 0
+
+    # --- Connection power queries ---
+
     def connection_power(self) -> HighspyArray | NDArray[Any]:
```
TrentHouliston, line 296: This should be the tagged version here. If it's needed at all anymore

custom_components/haeo/core/model/element.py
```diff
@@ -151,14 +202,107 @@ def connection_power(self) -> HighspyArray | NDArray[Any]:
...
             if end == "source":
-                # Power flowing into this element (as source)
                 total_power = total_power + conn.power_into_source
             elif end == "target":
-                # Power flowing into this element (as target)
                 total_power = total_power + conn.power_into_target
 
         return total_power
 
+    def connection_power_for_tag(self, tag: int) -> HighspyArray | NDArray[Any]:
```
TrentHouliston, line 320: Don't make specific tagged versions. We just have one version which is the tagged version (don't name it tag as if we put tag on everything, we may as well put tag on nothing)

TrentHouliston (commented):
I see a lot of this PR still straddling the line between using tags and having code paths that don't use tags. We will commit to using tags, and the "untagged" version is just a single tag for everything.

docs/modeling/tagged-power.md
```diff
@@ -0,0 +1,143 @@
...
+| --------- | ------------ | ------------- | ---------------------------- |
+| Solar     | 1            | —             | Source only, produces tag 1  |
+| Battery   | 2            | {1, 2, 3}    | Produces tag 2, consumes all |
+| Grid      | 3            | {1, 2, 3}    | Produces tag 3, consumes all |
+| Load      | —            | {1, 2, 3}    | Sink only, consumes all      |
+
+With `tag_costs` on the grid export connection:
+
+- Solar export (tag 1): +$0.02/kWh surcharge
+- Battery export (tag 2): +$0.10/kWh surcharge
```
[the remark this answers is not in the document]
TrentHouliston: :robot: Applied mdformat to all doc files.

docs/modeling/model-layer/connections/connection.md
```diff
@@ -40,12 +40,14 @@ Each segment receives a `power_in` expression and exposes a `power_out` expressi
...
-| `periods`  | Time period durations (hours)                              |
-| `segments` | Ordered mapping of segment names to segment specifications |
+| Parameter    | Description                                                |
+| ------------ | ---------------------------------------------------------- |
+| `source`     | Name of the source element                                 |
+| `target`     | Name of the target element                                 |
+| `periods`    | Time period durations (hours)                              |
+| `segments`   | Ordered mapping of segment names to segment specifications |
+| `tags`       | Set of tag IDs for per-source power decomposition (see [Tagged Power](../../tagged-power.md)) |
+| `tag_costs`  | Per-tag cost adjustments (list of `{tag, price}` dicts)    |
```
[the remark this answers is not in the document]
TrentHouliston: :robot: Applied mdformat to all doc files.

docs/modeling/tagged-power.md
```diff
@@ -0,0 +1,143 @@
...
+Consumption is distributed across the element's `access_list` tags.
+
+### Tag parameters
+
+| Parameter     | Scope      | Description                                                    |
+| ------------- | ---------- | -------------------------------------------------------------- |
+| `source_tag`  | Element    | Tag carried by power produced at this element                  |
+| `access_list` | Element    | Set of tags this element can consume (None = all)              |
+| `tags`        | Connection | Set of tags carried on this connection                         |
+| `tag_costs`   | Connection | Per-tag cost adjustments (`{tag, price}` entries)              |
```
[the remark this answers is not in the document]
TrentHouliston: :robot: Applied mdformat to all doc files.

docs/modeling/tagged-power.md
```diff
@@ -0,0 +1,143 @@
...
+## Example
+
+Consider a system with solar, battery, grid, and load connected through a switchboard:
+
+| Element   | `source_tag` | `access_list` | Role                         |
+| --------- | ------------ | ------------- | ---------------------------- |
+| Solar     | 1            | —             | Source only, produces tag 1  |
+| Battery   | 2            | {1, 2, 3}    | Produces tag 2, consumes all |
+| Grid      | 3            | {1, 2, 3}    | Produces tag 3, consumes all |
+| Load      | —            | {1, 2, 3}    | Sink only, consumes all      |
```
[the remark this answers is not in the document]
TrentHouliston: :robot: Applied mdformat to all doc files.

docs/modeling/model-layer/elements/node.md
```diff
@@ -23,18 +23,22 @@ This design separates the element's role (source, sink, or junction) from its op
...
 A pure junction (`is_source=false, is_sink=false`) creates no power variables.
 
 ### Parameters
 
-Node has no direct parameters.
-All operational parameters (power limits, efficiency, pricing) are configured on the connected [Connection](../connections/index.md) elements.
+| Parameter     | Type             | Description                                                               |
+| ------------- | ---------------- | ------------------------------------------------------------------------- |
+| `source_tag`  | `int \| None`    | Tag assigned to power produced by this element (see [Tagged Power](../../tagged-power.md)) |
+| `access_list` | `list[int] \| None` | Tags this element can consume (None = all tags)                        |
```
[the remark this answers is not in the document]
TrentHouliston: :robot: Applied mdformat to all doc files.

docs/modeling/model-layer/elements/node.md
```diff
@@ -23,18 +23,22 @@ This design separates the element's role (source, sink, or junction) from its op
...
 For each time step $t \in \{0, 1, \ldots, T-1\}$:
 
-| Variable            | Created When     | Domain                | Description                      |
-| ------------------- | ---------------- | --------------------- | -------------------------------- |
-| $P_{\text{in}}(t)$  | `is_sink=true`   | $\mathbb{R}_{\geq 0}$ | Power accepted from network (kW) |
-| $P_{\text{out}}(t)$ | `is_source=true` | $\mathbb{R}_{\geq 0}$ | Power provided to network (kW)   |
+| Variable                   | Created When     | Domain                 | Description                      |
+| -------------------------- | ---------------- | ---------------------- | -------------------------------- |
+| $P_{\text{produced}}(t)$   | `is_source=true` | $\mathbb{R}_{\geq 0}$ | Power provided to network (kW)   |
+| $P_{\text{consumed}}(t)$   | `is_sink=true`   | $\mathbb{R}_{\geq 0}$ | Power accepted from network (kW) |
```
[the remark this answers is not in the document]
TrentHouliston: :robot: Applied mdformat to all doc files.

custom_components/haeo/core/model/network.py
TrentHouliston: :robot: Reverted. `kModelEmpty` removed — only `kOptimal` is accepted now. An empty model should be an error.

custom_components/haeo/core/model/elements/node.py
TrentHouliston: :robot: Done. Node now always creates LP variables for both `_produced` and `_consumed`. Non-source nodes get zero-bounded production variables (`lb=0, ub=0`); non-sink nodes get zero-bounded consumption variables. `element_power_produced()` and `element_power_consumed()` always return HighspyArray — no None checks needed.

custom_components/haeo/core/model/element.py
TrentHouliston: :robot: Removed all getter/setter boilerplate. `outbound_tags` and `inbound_tags` are now plain instance variables set in `__init__`.

docs/modeling/model-layer/connections/connection.md
TrentHouliston: :robot: Updated connection.md to link to the segments index page instead of listing specific segment types inline. The connection doc now says "See segments for details on each type."

docs/modeling/model-layer/connections/connection.md
TrentHouliston: :robot: Fully committed to tagged power throughout. All "when tags are provided" language removed from docs and code. Connections now always have at least one tag (default `{0}` when none specified). Per-tag LP variables are always created. No dual code paths.

custom_components/haeo/core/model/tests/test_data/node.py
TrentHouliston: :robot: Changed from `is_source=False` to `is_source=True` because `element_power_balance` (from the Element base class) only creates constraints when there's non-zero produced/consumed power or connections. A non-source node with no connections returned int 0 for production, causing the balance constraint to be skipped — and thus no shadow price output to test. Making it a source ensures the balance constraint exists so the shadow price can be verified.

custom_components/haeo/core/model/elements/segments/segment.py
TrentHouliston: :robot: Done. Segments now only receive `power_in: dict[int, HighspyArray]` — always per-tag, no separate `tag_flows_in`. The `power_out` property also returns `dict[int, HighspyArray]`. Helper properties `total_power_in`/`total_power_out` sum the dict values for cases that need the aggregate.

custom_components/haeo/core/model/elements/segments/__init__.py
TrentHouliston: :robot: Done. `create_segment` now takes `power_in: dict[int, HighspyArray]` — no separate `tag_flows_in` parameter. The tagged flow dict is the only input.

custom_components/haeo/core/model/elements/segments/efficiency.py
TrentHouliston: :robot: Done. EfficiencySegment takes `power_in: dict[int, HighspyArray]` and `power_out` returns `{tag: flow * efficiency}`. No separate tagged/untagged paths.

custom_components/haeo/core/model/elements/segments/passthrough.py
TrentHouliston: :robot: Done. PassthroughSegment takes `power_in: dict[int, HighspyArray]`. Only tagged flows.

custom_components/haeo/core/model/element.py
TrentHouliston: :robot: Done. Renamed `source_tag` → `outbound_tags` (set[int] | None) and `access_list` → `inbound_tags` (set[int] | None) throughout the codebase. Both are now sets, giving a consistent interface. `None` means "all tags".

custom_components/haeo/core/model/elements/battery.py
TrentHouliston: :robot: Fixed. `BATTERY_POWER_CONSTRAINTS` is now `Final[frozenset[BatteryConstraintName]]` instead of `Final[frozenset[str]]`.

custom_components/haeo/core/model/elements/connection.py
TrentHouliston: :robot: Removed all separate tagged/untagged variants. `_initialize_segments` always creates per-tag LP variables and passes dicts through the segment chain. `power_into_source_for_tag`/`power_into_target_for_tag` look up directly from the dict. No conditional paths.

custom_components/haeo/core/model/elements/connection.py
TrentHouliston: :robot: Done. `_power_in` and `_power_out` on Connection are now `dict[int, HighspyArray]` only. `power_in`/`power_out` properties return the dict. `total_power_in`/`total_power_out` provide the sum for node power balance. No HighspyArray-based alternatives.

custom_components/haeo/core/model/elements/segments/pricing.py
TrentHouliston: :robot: Done. Simplified to a list comprehension:
```python
return Highs.qsum([
    self._power_in[tag] * price * self._periods
    for tag, price in self._tag_prices.items()
    if tag in self._power_in
])
```

custom_components/haeo/core/model/elements/connection.py
TrentHouliston: :robot: Removed. `_tag_power_in`/`_tag_power_out` are gone. The primary `_power_in`/`_power_out` dicts are the only storage.

custom_components/haeo/core/model/elements/segments/pricing.py
TrentHouliston: :robot: Done. Renamed `tag_costs` → `tag_prices` in `PricingSegmentSpec` and `PricingSegment`. The field is a per-unit rate (price), not an absolute amount (cost).

custom_components/haeo/core/model/elements/connection.py
TrentHouliston: :robot: Done. Removed all "when tags are provided" language from connection.py docstring. The connection always creates per-tag LP variables.

custom_components/haeo/core/model/tests/test_element.py
TrentHouliston: :robot: Done. Now uses `ELEMENT_POWER_BALANCE` constant instead of the literal string `"element_power_balance"` in the test assertions.

custom_components/haeo/core/model/elements/node.py
TrentHouliston: :robot: Removed. Node docstring now describes the four source/sink behavior modes concisely without referencing the produced/consumed bounds implementation detail that belonged to the old node_power_balance. The behavior is now encoded in variable bounds (zero-bounded for non-source/non-sink) with the Element base class handling power balance.

custom_components/haeo/core/model/element.py
TrentHouliston: :robot: Done. Merged `element_power_balance` and `element_tag_balance` into a single `element_power_balance` that always handles per-tag decomposition. When `connection_tags()` returns tags, it does per-tag balance with `outbound_tags`/`inbound_tags` routing. When there are no tags (no connections), it falls back to the simple total balance or skips entirely.

custom_components/haeo/core/model/element.py
TrentHouliston: :robot: Done. No more separate tagged version. `element_power_balance` is the single constraint method that always decomposes by tag when tags exist. The `element_tag_balance` method is gone.

custom_components/haeo/core/model/elements/node.py
TrentHouliston: :robot: Fixed. `NODE_POWER_BALANCE` is now `Final[NodeOutputName]` (with a type: ignore[assignment] for the literal narrowing) instead of `Final[str]`.

docs/modeling/model-layer/connections/connection.md
```diff
@@ -18,34 +18,34 @@ The mapping keys become segment names and drive the nested `segments` output.
...
-| `periods`  | Time period durations (hours)                              |
-| `segments` | Ordered mapping of segment names to segment specifications |
+| Parameter    | Description                                                                                   |
+| ------------ | --------------------------------------------------------------------------------------------- |
+| `source`     | Name of the source element                                                                    |
+| `target`     | Name of the target element                                                                    |
+| `periods`    | Time period durations (hours)                                                                 |
+| `segments`   | Ordered mapping of segment names to segment specifications                                    |
+| `tags`       | Set of tag IDs for per-source power decomposition (see [Tagged Power](../../tagged-power.md)) |
+| `tag_prices` | Per-tag price adjustments (list of `{tag, price}` dicts)                                      |
```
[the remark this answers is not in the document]
TrentHouliston: :robot: Fixed. Removed `tag_prices` from the Connection parameters table — it's a pricing segment parameter, not a connection-level one.

custom_components/haeo/core/model/element.py
```diff
@@ -140,25 +176,105 @@ def connection_power(self) -> HighspyArray | NDArray[Any]:
...
+
+        # Sum constraint: consumed_by_tag values must equal total consumed
+        if consumed_by_tag:
+            constraints.extend(list(reduce(operator.add, consumed_by_tag.values()) == consumed))
+
+        for tag in sorted(tags):
+            conn_tag = self.connection_power_for_tag(tag)
+            tag_prod = produced if tag in outbound else 0
+            tag_cons = consumed_by_tag.get(tag, 0)
+            constraints.extend(list(conn_tag + tag_prod - tag_cons == 0))
```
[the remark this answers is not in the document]
TrentHouliston: :robot: Valid observation. When `outbound` contains multiple tags, the full `produced` expression appears in each tag's balance equation, effectively duplicating production. For consumption we correctly decompose into `consumed_by_tag` variables with a sum constraint. Production should get the same treatment for multi-tag outbound. Currently all elements use single outbound tags (solar=1, battery=2, grid=3), so this doesn't manifest. Noted for when multi-tag production is needed — the fix would be mirroring `consumed_by_tag` with a `produced_by_tag` decomposition.

custom_components/haeo/core/model/element.py
```diff
@@ -36,6 +41,8 @@ def __init__(
         *,
         solver: Highs,
         output_names: frozenset[OutputNameT],
+        outbound_tags: set[int] | None = None,
+        inbound_tags: set[int] | None = None,
```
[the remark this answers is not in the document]
TrentHouliston: :robot: Acknowledged. The PR description still references the original `source_tag`/`access_list` names which were renamed to `outbound_tags`/`inbound_tags` during review. Will update the PR description to match the current API.

custom_components/haeo/core/model/tests/test_policy_scenario.py
```diff
@@ -0,0 +1,225 @@
...
+
+    assert isinstance(grid_imp, Connection)
+    assert isinstance(grid_exp, Connection)
+    assert isinstance(bat_dis, Connection)
+
+    grid_import_flow = tuple(float(v) for v in h.vals(grid_imp.total_power_in))
+    grid_export_flow = tuple(float(v) for v in h.vals(grid_exp.total_power_in))
+    bat_discharge_flow = tuple(float(v) for v in h.vals(bat_dis.total_power_in))
+
+    # Period 1: Solar surplus, cheap export ($0.01) + tag cost $0.02 = negative
```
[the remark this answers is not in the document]
TrentHouliston: :robot: Fixed. Comment now reads "positive net cost" — the math is $0.01 export + $0.02 tag surcharge = $0.03 positive, which explains why export is ~0.

custom_components/haeo/core/model/element.py
```diff
@@ -140,25 +176,105 @@ def connection_power(self) -> HighspyArray | NDArray[Any]:
...
+
+        produced = self.element_power_produced()
+        consumed = self.element_power_consumed()
+
+        # Determine which tags can carry produced/consumed power
+        outbound = set(tags) if self.outbound_tags is None else (self.outbound_tags & tags)
+        inbound = set(tags) if self.inbound_tags is None else (self.inbound_tags & tags)
+
+        # Get or create per-tag consumption variables
+        has_consumption = not (isinstance(consumed, (int, float)) and consumed == 0)
```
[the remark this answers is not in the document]
TrentHouliston: :robot: Valid edge case. If `inbound_tags` has no overlap with `connection_tags()`, the `consumed_by_tag` dict is empty, the sum constraint is skipped, and the consumption variable is unconstrained in the power balance. In practice, the optimizer would minimize it to zero (consumption has a cost or is bounded), but it's not explicitly enforced. This would represent a configuration error (element can consume but no connections carry the right tags). Noted as a guard to consider if we add config-time validation for tag consistency.

docs/modeling/tagged-power.md
```diff
@@ -0,0 +1,143 @@
...
+Consumption draws from the element's `inbound_tags`.
+
+### Tag parameters
+
+| Parameter       | Scope      | Description                                            |
+| --------------- | ---------- | ------------------------------------------------------ |
+| `outbound_tags` | Element    | Tags that produced power can be placed on (None = all) |
+| `inbound_tags`  | Element    | Tags that consumed power can draw from (None = all)    |
+| `tags`          | Connection | Set of tags carried on this connection                 |
+| `tag_prices`    | Connection | Per-tag price adjustments (`{tag, price}` entries)     |
```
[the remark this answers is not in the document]
TrentHouliston: :robot: Fixed. Updated the scope from "Connection" to "Pricing segment" in the tag parameters table.

custom_components/haeo/core/model/elements/connection.py
```diff
@@ -128,24 +137,45 @@ def _initialize_segments(self, source_element: Element[Any], target_element: Ele
...
         return self._power_in
 
     @property
-    def power_out(self) -> HighspyArray:
-        """Power exiting the connection at the target end (after transforms)."""
-        assert self._power_out is not None  # noqa: S101
+    def total_power_in(self) -> HighspyArray:
+        """Total power entering the connection (sum of all tags)."""
+        return reduce(operator.add, self._power_in.values())
+
```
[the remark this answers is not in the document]
TrentHouliston: :robot: Not applicable. Same as the `total_power_out` comment — `_power_in` is always initialized by `set_endpoints()` during `Network.add()`, which is the only construction path. Per project conventions, we rely on errors for coding bugs rather than adding defensive checks.

custom_components/haeo/core/model/element.py
```diff
@@ -1,15 +1,19 @@
...
-from highspy.highs import HighspyArray, highs_cons
+from highspy.highs import HighspyArray, highs_cons, highs_linear_expression
 import numpy as np
 from numpy.typing import NDArray
 
 from .output_data import OutputData
-from .reactive import OutputMethod, ReactiveConstraint, ReactiveCost, TrackedParam, cost
+from .reactive import OutputMethod, ReactiveConstraint, ReactiveCost, TrackedParam, constraint, cost
+
+ELEMENT_POWER_BALANCE: Final = "element_power_balance"
```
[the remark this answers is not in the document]
TrentHouliston: :robot: Not applicable. Pyright already infers `Literal["element_power_balance"]` from bare `Final` — adding `Final[Literal[...]]` triggers ruff PYI064. The `type: ignore[assignment]` on `NODE_POWER_BALANCE` was a pre-existing artifact and has been removed; bare `Final` works correctly.

custom_components/haeo/core/model/element.py
```diff
@@ -140,25 +176,105 @@ def connection_power(self) -> HighspyArray | NDArray[Any]:
...
+        # Sum constraint: consumed_by_tag values must equal total consumed
+        if consumed_by_tag:
+            constraints.extend(list(reduce(operator.add, consumed_by_tag.values()) == consumed))
+
+        for tag in sorted(tags):
+            conn_tag = self.connection_power_for_tag(tag)
+            tag_prod = produced if tag in outbound else 0
+            tag_cons = consumed_by_tag.get(tag, 0)
+            constraints.extend(list(conn_tag + tag_prod - tag_cons == 0))
+
```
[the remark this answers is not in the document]
TrentHouliston: :robot: Valid observation. The current constraint `connection_power_for_tag(tag) == 0` enforces net-zero across all connections at the element, which does allow pass-through (inflow on one connection balanced by outflow on another). True blocking would require per-connection per-tag constraints forcing each individual flow to zero. For the current use case (single outbound tag per element), this distinction doesn't arise. Noted as a design consideration for when multi-tag routing is exercised more fully.

custom_components/haeo/core/model/elements/connection.py
```diff
@@ -128,24 +137,45 @@ def _initialize_segments(self, source_element: Element[Any], target_element: Ele
...
+    @property
+    def power_out(self) -> dict[int, HighspyArray]:
+        """Per-tag power exiting the connection at the target end."""
         return self._power_out
 
+    @property
+    def total_power_out(self) -> HighspyArray:
+        """Total power exiting the connection (sum of all tags)."""
+        return reduce(operator.add, self._power_out.values())
```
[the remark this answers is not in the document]
TrentHouliston: :robot: Not applicable. Per project conventions: "Always assume that accessed properties/fields which should exist do exist directly. Rely on errors occurring if they do not when they indicate a coding error." `_power_out` is always initialized before `total_power_out` is accessed — `set_endpoints()` is called by `Network.add()` during connection registration, which is the only construction path.

custom_components/haeo/core/model/element.py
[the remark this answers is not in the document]
TrentHouliston: :robot: Fixed in 3d9a42c. When `inbound` is empty (no overlap between `inbound_tags` and `connection_tags()`) but the element has consumption, a `consumed == 0` constraint is now added. Same guard for `outbound`/production. Added TDD test `test_empty_inbound_forces_consumption_to_zero`.

custom_components/haeo/core/model/element.py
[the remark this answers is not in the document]
TrentHouliston: :robot: Fixed in 3d9a42c. Production is now decomposed into per-tag `produced_by_tag` LP variables with a sum constraint (`Σ produced_by_tag == produced`), mirroring the existing `consumed_by_tag` pattern. Each outbound tag gets its own share of production. Added TDD test `test_multi_tag_outbound_does_not_duplicate_production`.

custom_components/haeo/core/model/element.py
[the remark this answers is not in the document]
TrentHouliston: :robot: Fixed in 3d9a42c. Blocked tags now force each connection's per-tag flow to zero individually (`conn.power_into_source_for_tag(tag) == 0` and `conn.power_into_target_for_tag(tag) == 0`) instead of only constraining the net sum. This prevents pass-through of blocked tags across multiple connections. Added TDD test `test_blocked_tag_forces_per_connection_flow_to_zero`.

PR #374, opened 2026-04-16
Policy config flow UX fixes

TrentHouliston:
Five UX issues in the policy config flow introduced by the power policy system.

### Singular "Policy" for adding
The `entry_type` and `initiate_flow.user` translations said "Policies" but should say "Policy" since the user is adding a single rule. Updated translations while keeping the subentry title as "Policies".

### Edit rule shows previous values
`_rule_to_defaults()` returned raw values (`["Solar"]` for endpoints, `{"type": "constant", "value": 0.02}` for price) instead of the ChooseSelector format expected by `add_suggested_values_to_schema`. Now returns:
- Source/target: `{"active_choice": "nodes", "nodes": [...]}` for specific elements, `""` for any
- Price: raw float for constants, entity ID list for entities, `""` for none

### "Nodes" renamed to "Elements"
The selector choice label "Nodes" was confusing since policies can target any element type. Changed to "Elements".

### Enable/disable toggle per rule
Added a `BooleanSelector` for `enabled` field on each rule. Disabled rules are stored in config but skipped by `extract_policy_rules()` in the adapter.

### Empty element selection normalizes to "any"
Selecting "Elements" in the endpoint chooser but leaving the multi-select empty now normalizes to "any" (empty string) instead of storing an empty list.

## Testing

8 new tests covering all behavioral changes (edit defaults format, enabled/disabled storage and filtering, empty nodes normalization). All 54 tests pass.

### DIFF
--- a/custom_components/haeo/flows/elements/policy.py
+++ b/custom_components/haeo/flows/elements/policy.py
@@ -13,6 +13,7 @@ from typing import Any
 
 from homeassistant.config_entries import ConfigSubentry, ConfigSubentryFlow, SubentryFlowResult
 from homeassistant.helpers.selector import (
+    BooleanSelector,
     ChooseSelector,
     ChooseSelectorChoiceConfig,
     ChooseSelectorConfig,
@@ -26,9 +27,10 @@ from homeassistant.helpers.selector import (
 import voluptuous as vol
 
 from custom_components.haeo.core.const import CONF_ELEMENT_TYPE, CONF_NAME
-from custom_components.haeo.core.schema.constant_value import as_constant_value
+from custom_components.haeo.core.schema.constant_value import as_constant_value, is_constant_value
 from custom_components.haeo.core.schema.elements.element_type import ElementType
 from custom_components.haeo.core.schema.elements.policy import (
+    CONF_ENABLED,
     CONF_PRICE,
     CONF_RULE_NAME,
     CONF_RULES,
@@ -37,8 +39,8 @@ from custom_components.haeo.core.schema.elements.policy import (
     ELEMENT_TYPE,
     PolicyRuleConfig,
 )
-from custom_components.haeo.core.schema.entity_value import as_entity_value
-from custom_components.haeo.core.schema.none_value import as_none_value
+from custom_components.haeo.core.schema.entity_value import as_entity_value, is_entity_value
+from custom_components.haeo.core.schema.none_value import as_none_value, is_none_value
 from custom_components.haeo.elements import get_list_input_fields
 from custom_components.haeo.elements.input_fields import InputFieldInfo
 from custom_components.haeo.flows.element_flow import ElementFlowMixin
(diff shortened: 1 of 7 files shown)


custom_components/haeo/flows/elements/policy.py
```diff
@@ -167,6 +172,7 @@ def _build_rule_schema(self, participants: list[str]) -> vol.Schema:
 
         return vol.Schema(
             {
+                vol.Optional(CONF_ENABLED, default=True): BooleanSelector(),
```
TrentHouliston: put enabled after the name

custom_components/haeo/flows/elements/policy.py
```diff
@@ -227,12 +236,25 @@ def _rule_to_defaults(self, rule: PolicyRuleConfig) -> dict[str, Any]:
         defaults: dict[str, Any] = {
             CONF_RULE_NAME: rule["name"],
         }
+        if CONF_ENABLED in rule:
+            defaults[CONF_ENABLED] = rule[CONF_ENABLED]
+
         if "source" in rule:
```
TrentHouliston: These feel like ternaries

custom_components/haeo/flows/tests/test_policy_flows.py
```diff
@@ -1,6 +1,7 @@
 """Tests for the policy element config flow."""
 
 from types import MappingProxyType
+from typing import Any
```
TrentHouliston, line 4: Is any needed? I'm always suspicous that means that there is something wrong with the types

custom_components/haeo/flows/elements/policy.py
```diff
@@ -74,7 +76,10 @@ def __call__(self, data: Any) -> Any:
             if choice == CHOICE_NONE:
                 return ""
             if choice == CHOICE_NODES:
-                return data.get(CHOICE_NODES, [])
+                nodes = data.get(CHOICE_NODES, [])
+                if not nodes:
+                    return ""
+                return nodes
```
TrentHouliston: This seems like a very complicated way of just returning an empty string, an empty array is falsey already right?

custom_components/haeo/flows/elements/policy.py
```diff
@@ -227,12 +236,25 @@ def _rule_to_defaults(self, rule: PolicyRuleConfig) -> dict[str, Any]:
         defaults: dict[str, Any] = {
             CONF_RULE_NAME: rule["name"],
         }
+        if CONF_ENABLED in rule:
```
TrentHouliston: It should just be a required field that defaults to true

custom_components/haeo/flows/elements/policy.py
TrentHouliston: Also shouldn't these be constants not literals?

custom_components/haeo/translations/en.json
```diff
@@ -588,7 +588,7 @@
       "options": { "2_days": "2 days", "3_days": "3 days", "5_days": "5 days", "7_days": "7 days", "custom": "Custom" }
     },
     "input_source": { "choices": { "constant": "Constant", "entity": "Entity", "none": "None" } },
-    "policy_endpoint": { "choices": { "nodes": "Nodes", "none": "Any" } }
+    "policy_endpoint": { "choices": { "nodes": "Elements", "none": "Any" } }
```
TrentHouliston: You still have nodes as the key here, should that be elements and change it where it needs to change?

TrentHouliston (commented):
I'm not 100% sure if it is or not. But make sure that the enabled flag creates an entity on the device that can be toggled and is synced with the configuration the same as how the numbers are.

TrentHouliston:
There should also be a vol validation you can do that will prevent the user from multiselecting and having 0 elements, some sort of length 1 check

PR #375, opened 2026-04-16
Fix optimizer re-run race condition in async_update_subentry_value

TrentHouliston:
Fixes a race condition where value updates to input entities triggered a full integration reload instead of a lightweight optimizer re-run.

## Problem

`async_update_subentry_value()` set `value_update_in_progress = True`, called `async_update_subentry`, then immediately cleared the flag in a `finally` block. However, `async_update_subentry` dispatches update listeners via `hass.async_create_task()` — they run in the **next** event loop tick. The flag was already cleared before the listener could see it, so `async_update_listener` fell through to `async_reload()` instead of `signal_optimization_stale()`.

## Fix

Remove the `try/finally` block. The update listener in `__init__.py:async_update_listener` already handles clearing `value_update_in_progress` after checking it:

```python
if runtime_data and runtime_data.value_update_in_progress:
    runtime_data.value_update_in_progress = False  # ← listener clears it
    coordinator.signal_optimization_stale()
    return
```

The function now sets the flag and calls the update — the listener is responsible for clearing it.

## Testing

- New test `test_flag_remains_set_after_call` verifies the flag stays `True` after the function returns
- Updated existing assertion in `test_editable_mode_set_native_value_with_runtime_data` to expect `True`
- All 56 tests pass

### DIFF
--- a/custom_components/haeo/util/__init__.py
+++ b/custom_components/haeo/util/__init__.py
@@ -41,16 +41,11 @@ async def async_update_subentry_value(
     new_data = dict(subentry.data)
     set_nested_config_value_by_path(new_data, field_path, value)
 
-    try:
-        hass.config_entries.async_update_subentry(
-            entry,
-            subentry,
-            data=new_data,
-        )
-    finally:
-        # Ensure flag is cleared even if update fails
-        if runtime_data is not None:
-            runtime_data.value_update_in_progress = False
+    hass.config_entries.async_update_subentry(
+        entry,
+        subentry,
+        data=new_data,
+    )
 
 
 __all__ = [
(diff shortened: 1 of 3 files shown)


custom_components/haeo/util/__init__.py
```diff
@@ -41,16 +41,11 @@ async def async_update_subentry_value(
     new_data = dict(subentry.data)
     set_nested_config_value_by_path(new_data, field_path, value)
 
-    try:
-        hass.config_entries.async_update_subentry(
-            entry,
-            subentry,
-            data=new_data,
-        )
-    finally:
```
TrentHouliston, line 50: Is there any concern with this change that a failure in the async call will result in the flag getting stuck and killing the optimisation forever?

PR #395, opened 2026-04-20
Encourage flow via strictly-negative secondary time-preference weights

TrentHouliston:
## Summary

- Change the per-connection time-preference secondary weights from strictly positive to strictly negative while preserving the existing priority / time ordering. Minimising the secondary now *rewards* carrying flow rather than penalising it.
- Plumb `priority_total` alongside `priority` from `Network.optimize` to each `Connection` so every connection normalises against the same global range.
- Update scenario snapshots (1–6); scenario6 now shows the battery charging from surplus solar, matching the old `early_charge_incentive` behaviour without needing a dedicated primary cost.

## Why

Previously the secondary objective was `w = priority*T + (t+1)` — all positive — so minimising it at cost parity *reduced* total flow. That meant free surplus solar was curtailed rather than stored whenever no downstream revenue justified charging. The new formulation keeps the same tie-break ordering (lower priority first, earlier period first) but shifts every weight below zero, so any flow improvement strictly benefits the secondary objective.

New weight formula per period `t` with priority `p` in a network with `P` connections over `T` periods:

```
w[p, t] = p*T + (t + 1) - (P*T + 1)      ∈ [-P*T, -1]
```

The largest weight is `-1`, the smallest is `-P*T`, and the ordering is identical to before.

## Notes on unboundedness

Networks with *zero* primary cost and *no* capacity segments become unbounded under the new formulation (flow goes to infinity to chase `-∞` secondary). This is extremely rare in practice — batteries, grids, solar and loads all impose primary costs or capacity limits. Two existing unit tests that constructed pathological zero-cost / unbounded topologies were updated to include either a positive price (0.10 \$/kWh) or a `power_limit` segment.

## Test plan

- [x] `uv run pytest custom_components/haeo` — 1577 passed
- [x] `uv run pytest tests/scenarios/ -m scenario` — 6 passed, snapshots regenerated
- [x] `uv run ruff check && uv run ruff format --check`
- [x] `uv run pyright`
- [x] `uv run lint-imports`
- [x] `npx prettier@3 --check .`
- [x] New regression tests:
  - `test_connection_secondary_weights_strictly_negative` — proves every per-period weight `< 0` and matches the expected formula.
  - `test_connection_secondary_preserves_time_order` — earlier periods fill before later ones.
  - `test_connection_secondary_respects_priority` — lower-priority connections fill before higher-priority ones.
  - `test_secondary_encourages_free_flow_into_battery` — battery absorbs free solar when no other sink benefits.


TrentHouliston:
This didn't work, it constantly exploded as having negative weights means it always wants to make power go in circles

PR #398, opened 2026-04-20
Encourage flow with asymmetric sink/source time-preference weights

TrentHouliston:
## Summary

Replaces the strictly-positive time-preference tie-breaker on every
connection with an **asymmetric signed weight** applied only at network
terminals (connections that touch a genuine source or sink).

- **Sink-terminal** connections (`target.is_sink`) get a
  **strictly-negative** weight — early consumption is rewarded, so free
  upstream power (e.g. solar) actively pulls into loads and storage
  even when no primary-cost incentive exists.
- **Source-terminal** connections (`source.is_source`) get a
  **strictly-positive** weight scaled by `SOURCE_PENALTY_FACTOR` (0.5)
  — early production is mildly penalised, so stored energy is held in
  reserve unless a sink on the other end of the path rewards the flow.
- **Non-terminal transfer** connections (e.g. an inverter's DC↔AC
  internals) contribute **no** secondary term, so the solver cannot
  saturate zero-primary-cost loops just to improve the tie-breaker.

### Why

This fixes the regression observed on `scenario6`: with the battery's
sell price set above the grid's buy price (so round-tripping to sell is
unprofitable), the battery refused to charge from free solar — it
either left energy on the table or, under a universally-negative
variant of the tie-breaker, circulated power through the inverter to
consume the secondary "budget" without storing anything.

The asymmetric design also **dampens phantom round-tripping** at a
single element: charge + discharge in the same period picks up a
negative sink weight *and* a positive source weight which partly
cancel, so the tie-breaker no longer pays the solver to cycle a
battery against itself.

### Supporting changes

- `NetworkElement.is_source` / `is_sink` are now first-class metadata
  (defaults `False`); `Battery` overrides both to `True`.
- `Connection` exposes `is_terminal_sink`, `is_terminal_source`, and
  `is_terminal` properties driven by the endpoint metadata.
- `Network._assign_priorities` propagates a shared `priority_total` so
  every connection normalises its weights against the same global
  range, keeping magnitudes comparable and the sign guarantee intact.
- Two `Network`-level tests that previously relied on the *positive*
  secondary to bound a zero-primary-cost LP now include a non-zero
  price or an explicit power limit, because terminal sink-side
  weights are now negative and would otherwise let the LP run to
  infinity on the secondary alone.
- `scenario6`'s `Battery min sell` policy price was raised to `1.04`
  so the snapshot exercises the "free solar with no profitable sell
  path" case that originally surfaced the bug.
- All scenario snapshots (1–6) regenerated to reflect the new
  tie-breaker behaviour.
- Docs updated: `docs/modeling/index.md` and
  `docs/modeling/model-layer/connections/connection.md` describe the
  terminal-only application and the sink/source sign asymmetry.

## Test plan

- [x] `uv run ruff check` on all modified files.
- [x] `uv run pytest custom_components/haeo/core/model/tests/test_network.py` — all pass.
- [x] `uv run pytest custom_components/haeo/core/model/elements/tests/test_connections.py` — all pass (including new `test_sink_connection_has_strictly_negative_weights`, `test_source_connection_has_strictly_positive_weights`, `test_source_and_sink_connection_partially_cancels`, `test_round_trip_pair_reduces_secondary_reward`).
- [x] `uv run pytest` (full suite, 1728 passed, 2 skipped).
- [x] Visual review of regenerated `scenario6` SVG: battery now charges to 100% from free solar even when selling is unprofitable.


TrentHouliston:
Closing — after investigation, scenario 6's under-charging isn't a bug: it's the correct multi-day optimum given locked-capacity effects (useful discharge is bounded by load × night-length, so charge > ~20% is locked capacity with no productive outlet before the next solar day). The secondary objective isn't the right lever for this; encoding it as a small `-0.001` Solar→Battery policy (which is what v0.3.3 effectively had) is the cleaner fix. Follow-up PR will re-enable v1.3 migration of the legacy early-charge incentive to a real policy.

PR #399, opened 2026-04-20
Migrate legacy early-charge-incentive default through to a real policy

TrentHouliston:
## Summary

The v1.3 migration previously listed `0.001` in its
`_DEFAULT_POLICY_PRICES[(battery, price_target_source)]` set, which
caused the legacy v0.3.3 `early_charge_incentive` default value to be
silently dropped during migration instead of being converted into a
policy.

That small incentive was doing meaningful work on `main`: it was the
only thing making the optimiser prefer storing free upstream solar over
leaving the battery at the minimum SoC required to cover the next
night's load.  Without it, any install whose export path is profitable
(however marginally) now leaves excess solar un-stored, because the
secondary tie-breaker only breaks ties at equal primary cost — it
doesn't compete with real export revenue, however small.

Remove `0.001` from the default set so it migrates through as a real
`target=[Battery], price=-0.001` rule.  `0.0` and `0.01` stay on the
default list unchanged (they never carried the early-charge semantics
and are still legitimately skippable).

## Why this is the right place for the fix

Investigation on scenario 6 (in an earlier thread) established that the
"battery doesn't fully charge from free solar" behaviour is actually
the correct multi-day optimum given only the prices the solver can
see — useful discharge is bounded by `load × night-length`, so
charging past ~`night_load / capacity` is locked-up capacity with no
productive outlet before the next solar day.  A small user-configurable
preference (expressed as a policy) is the right mechanism for
encouraging "fill the battery first anyway"; it was already in the
pre-1.3 schema as `early_charge_incentive` and the migration just
needs to carry it through instead of dropping it.

## Caveat

This fixes **new** migrations from pre-1.3 entries.  Installs that
already migrated to 1.3 under the old filter lost the incentive at
that point and will need to manually re-add a `target=[Battery],
price=-0.001` policy rule to restore the old behaviour.  Writing a
retroactive 1.4 migration to add it back was not in scope here —
happy to follow up if desired.

## Test plan

- [x] `uv run ruff check` on modified files.
- [x] `uv run pytest custom_components/haeo/migrations` — 55 passed (new unit test `test_is_default_policy_price_preserves_legacy_early_charge_incentive` and new integration test `test_async_migrate_entry_preserves_legacy_early_charge_incentive` cover both the filter and the end-to-end rule generation).
- [x] `uv run pytest` (full suite) — 1721 passed, 2 skipped.

Closes the discussion that led to #398 (now closed).

### DIFF
--- a/custom_components/haeo/migrations/tests/test_v1_3.py
+++ b/custom_components/haeo/migrations/tests/test_v1_3.py
@@ -764,6 +764,43 @@ async def test_async_migrate_entry_keeps_colliding_section_unique_ids_distinct(h
     assert registry.async_get_entity_id("number", DOMAIN, f"{entry.entry_id}_bat001_partition_percentage") is None
 
 
+async def test_async_migrate_entry_preserves_legacy_early_charge_incentive(hass: HomeAssistant) -> None:
+    """Legacy 0.001 charge-incentive migrates to a -$0.001 target=Battery policy.
+
+    Previously the migration silently dropped 0.001 as a "default" value;
+    with no implicit early-charge behaviour in the new optimizer that
+    would change existing installs' pull-free-power-into-storage
+    preference.  The migration now carries it through as a real policy.
+    """
+    entry = MockConfigEntry(domain=DOMAIN, title="Hub", data={CONF_NAME: "Hub"}, version=1, minor_version=0)
+    entry.add_to_hass(hass)
+
+    battery_subentry = _create_subentry(
+        {
+            CONF_ELEMENT_TYPE: battery.ELEMENT_TYPE,
+            CONF_NAME: "Bat",
+            CONF_CONNECTION: "bus",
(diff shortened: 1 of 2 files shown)


purcell-lab:
Thanks,

I have written my own Early Charge Incentive, so I don't think retroactive 1.4 migrations are necessary, but good to have this as a standard migration.

image

nickw444:
> I have written my own Early Charge Incentive, so I don't think retroactive 1.4 migrations are necessary, but good to have this as a standard migration.

@purcell-lab I'm curious how you went about creating the incentive? Did you provide an entity with a forecast scaling down as the value for each time period? Or some other magic? Or is it simpler than that?

(I tried a sensor entity with a forecast scaling the value across the planning horizon, but policies only let you choose `switch` or `number` entities, and AFAICT `number` entities can't have attributes

PR #420, opened 2026-05-03
Fix battery undercharge partition acting as hard SOC clamp

sucitram:
## Summary

When `undercharge_percentage` exceeds `min_charge_percentage` (e.g. undercharge=20%, min=0%), the adapter anchored the LP coordinate origin at the undercharge boundary rather than the hardware minimum. This caused two related failures:

- **SOC pinned at partition level**: initial SOC values below the undercharge boundary were mapped to a negative `initial_charge`, clamped to zero by the LP model, and all forecasts showed SOC pinned at the partition level rather than the actual battery state.
- **Partition cost never fires**: `discharge_energy_threshold` was computed as `(min - undercharge) * capacity`, which is always ≤ 0 in this configuration. The soc_pricing slack variable was perpetually zero, so the undercharge penalty had no effect — the boundary acted as a hard floor regardless of the configured cost.

Both issues share the same root cause: the LP origin was in the wrong place.

## Changes

- **`battery.py` (adapter)**: anchor `lower_ratio` at `min(undercharge_percentage, min_charge_percentage)` so the LP origin always sits at the true hardware minimum. Compute `discharge_energy_threshold` as the distance from that origin to `max(undercharge, min_charge_percentage)`, which is always positive and activates the penalty at the correct SOC level. Apply the same minimum in `_calculate_total_energy` so displayed SOC is consistent with the model.
- **`test_battery_model.py`**: add parametrized cases covering the corrected `initial_charge` and SOC reconstruction when initial SOC falls inside the undercharge penalty zone. An `xfail` test documents the remaining edge case where initial SOC falls below the undercharge floor in an inverted configuration (`undercharge < min`), which requires model-layer changes to accept a negative `initial_charge`.

Configurations where `undercharge_percentage < min_charge_percentage` are unaffected: `min()` and `max()` return the same values as before, so all existing tests pass unchanged.

### DIFF
--- a/custom_components/haeo/core/adapters/elements/battery.py
+++ b/custom_components/haeo/core/adapters/elements/battery.py
@@ -125,7 +125,15 @@ class BatteryAdapter:
         undercharge_percentage = undercharge.get(CONF_PARTITION_PERCENTAGE) if undercharge_cost is not None else None
         overcharge_percentage = overcharge.get(CONF_PARTITION_PERCENTAGE) if overcharge_cost is not None else None
 
-        lower_ratio = undercharge_percentage if undercharge_percentage is not None else min_charge_percentage
+        # When undercharge_percentage > min_charge_percentage (e.g. user sets undercharge=20%,
+        # leaves min at default 0%), use min_charge_percentage as the LP origin so that
+        # (a) initial SOC values below the undercharge boundary remain representable without
+        #     clamping, and (b) the soc_pricing threshold is positive and can activate.
+        lower_ratio = (
+            np.minimum(undercharge_percentage, min_charge_percentage)
+            if undercharge_percentage is not None
+            else min_charge_percentage
+        )
         upper_ratio = overcharge_percentage if overcharge_percentage is not None else max_charge_percentage
 
         lower_ratio_first = float(lower_ratio[0]) if isinstance(lower_ratio, np.ndarray) else float(lower_ratio)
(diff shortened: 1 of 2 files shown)


purcell-lab:
Excellent, I had exactly this issue last night. Got to my undercharge %, stopped discharge and reverted to grid import.

TrentHouliston:
I think I really just need to overhaul how the partitions work.

Technically the undercharge is supposed to be below the minimum. I never really liked either naming though. You discharge to the "minimum" and undercharge below that is how it was supposed to be interpreted. But it's just as easy to put the undercharge as a limit and min as the absolute minimum.

So in this case I think it's that you had min below undercharge, so undercharge was taken as the absolute minimum and stopped there.

sucitram:
> Technically the undercharge is supposed to be below the minimum. I never really liked either naming though. You discharge to the "minimum" and undercharge below that is how it was supposed to be interpreted. But it's just as easy to put the undercharge as a limit and min as the absolute minimum.

In that case what I was running was simply a misconfiguration - my configuration should have been something more like:
- undercharge = 5% : hard floor (LP origin) - absolute minimum, cannot go below.
- min_charge =  20% : soft preferred floor - penalty activates when discharging below here.
- max_charge = 90% : soft preferred ceiling - penalty activates when charging above here.
- overcharge = 100% : hard ceiling (LP max) - absolute maximum, cannot exceed.

I added a couple of additional unit tests locally and the partitions appear to be working as designed.

PR #424, opened 2026-05-09
Expose shadow prices in $/kWh (closes #423)

### LINKED ISSUE #423: Please consider normalising shadow pricing to $/kWh
### Checklist

- [x] I have filled out the template to the best of my ability.
- [x] This only contains 1 feature request (if you have multiple feature requests, open one feature request for each feature request).
- [x] This issue is not a duplicate feature request of [previous feature requests](https://github.com/ludeeus/integration_blueprint/issues?q=is%3Aissue+label%3A%22Feature+Request%22+).

### Is your feature request related to a problem? Please describe.

Most shadow pricing currently exposed is in "$/kW", which then needs to be normalised for the interval duration to arrive at a more familiar $/kWh shadow price.

1 minute intervals need to be multiplied by 60, hourly intervals are multiplied by 1.

normalised shadow prices make it easier to implement rules like, schedule and run this device when the marginal cost is less than 0.15 $/kWh.

### Describe the solution you'd like

Provide shadow pricing, either updated existing, or a new set of sensors normalised to $/kWh.

### Describe alternatives you've considered

I currently have a workaround template sensor.

```yaml
  - sensor:
      - name: "switchboard_local_spot_price"
        unique_id: switchboard_local_spot_price
        unit_of_measurement: "$/kWh"
        state_class: measurement
        state: >-
          {%- set forecast = state_attr('sensor.switchboard_power_balance_shadow_price', 'forecast') -%}
          {%- if forecast is none or forecast | length < 2 -%}
            unavailable
          {%- else -%}
            {%- set t0 = (forecast[0].time | as_datetime).timestamp() -%}
            {%- set t1 = (forecast[1].time | as_datetime).timestamp() -%}
            {%- set interval_h = (t1 - t0) / 3600 -%}
            {{ (forecast[0].value | float / interval_h) | round(4) }}
          {%- endif -%}
        attributes:
          interval_minutes: >-
            {%- set forecast = state_attr('sensor.switchboard_power_balance_shadow_price', 'forecast') -%}
            {%- if forecast is none or forecast | length < 2 -%}
              unknown
            {%- else -%}
              {%- set t0 = (forecast[0].time | as_datetime).timestamp() -%}
              {%- set t1 = (forecast[1].time | as_datetime).timestamp() -%}
              {{ ((t1 - t0) / 60) | round(0) | int }}
            {%- endif -%}

          forecast: >-
            {%- set raw = state_attr('sensor.switchboard_power_balance_shadow_price', 'forecast') -%}
            {%- if raw is none or raw | length < 2 -%}
              []
            {%- else -%}
              {%- set ns = namespace(result=[]) -%}
              {%- for i in range(raw | length - 1) -%}
                {%- set t0 = (raw[i].time | as_datetime).timestamp() -%}
                {%- set t1 = (raw[i+1].time | as_datetime).timestamp() -%}
                {%- set interval_h = (t1 - t0) / 3600 -%}
                {%- set spot = (raw[i].value | float / interval_h) | round(4) -%}
                {%- set iso = (raw[i].time | as_datetime).isoformat() -%}
                {%- set ns.result = ns.result + [{'time': iso, 'value': spot}] -%}
              {%- endfor -%}
              {{ ns.result }}
            {%- endif -%}
```

### Additional context

Image

purcell-lab:
## Summary

Closes #423.

Every shadow-price sensor is now emitted in `$/kWh` so it sits on the same axis as tariffs and other energy-priced quantities. Instead of computing the duals in `$/kW` and converting to `$/kWh` in the adapter layer, this PR adopts @TrentHouliston's energy-native LP formulation (from [`experiment/energy-native-constraints`](https://github.com/hass-energy/haeo/compare/main...experiment/energy-native-constraints)): the LP-layer power-balance and power-limit constraints are multiplied by `dt` before being added to the solver, so HiGHS returns the duals in `$/kWh` directly. The adapter layer just passes them through.

This replaces the original PR scope (a `shadow_price_utils.py` helper + per-adapter division) with a much smaller and more principled change.

### Why the LP-layer approach is preferred

- **Less code.** Net 311 lines removed across the helper module, helper tests, adapter call sites, and stale documentation. The two LP-layer edits are 25 inserted / 15 deleted lines.
- **Dimensionally correct at the source.** Duals come out of HiGHS in `$/kWh`, with the unit baked into the constraint definition (`@constraint(output=True, unit="$/kWh")`). No risk of an adapter forgetting to divide.
- **Handles tagged/multi-period shadows naturally.** No more "is this value count a multiple of `len(periods)`?" branching in the helper.
- **Battery and battery-section shadows work uniformly.** Because `BATTERY_POWER_BALANCE := ELEMENT_POWER_BALANCE`, the dt-scaling carries over to battery elements without any extra adapter logic. The "battery deferred" caveat from the original PR is resolved.
- **Primal solution is unchanged.** Multiplying both sides of every constraint by `dt > 0` is a strict scaling, so the optimizer arrives at the same dispatch.

### Changes in this PR

LP layer (from Trent's experiment, cherry-picked with author preserved):

- `core/model/element.py` – `element_power_balance` constraint multiplied by `dt`.
- `core/model/elements/segments/power_limit.py` – both sides of the power-limit constraint multiplied by `dt`.
- Both constraints annotated `@constraint(output=True, unit="$/kWh")`.

Adapter cleanup:

- Removed `custom_components/haeo/core/adapters/shadow_price_utils.py` and its tests.
- Dropped `shadow_price_per_energy` calls and `numpy/NDArray` imports from `node`, `solar`, `load`, `grid`, `inverter`, `battery`, `battery_section` adapters.
- Simplified `outputs()` signatures: most adapters no longer accept `periods` (the coordinator still passes it keyword-only, absorbed by `**_kwargs`). `grid` keeps `periods` for its cost calculation.
- Decorator default unit flipped from `"$/kW"` → `"$/kWh"` in `reactive/decorators.py`.

Tests & docs:

- Adapter test fixtures updated to feed `$/kWh` inputs directly through the adapters (passthrough contract).
- `core/model/tests/test_data/{battery,connection,node}.py` unit strings updated.
- `docs/modeling/shadow-prices.md` rewritten to describe the energy-native LP formulation instead of the adapter-layer division.

### Sensor naming

All published sensors retain the `_shadow_energy_price` suffix and emit `$/kWh` (e.g. `node_power_balance_shadow_energy_price`, `grid_power_max_import_shadow_energy_price`, `inverter_dc_bus_shadow_energy_price`). The `EntityCategory.DIAGNOSTIC` tagging on shadow-price sensors is unchanged from the previous revision.

### Worked example (from #423)

For a 5-minute period (`dt = 1/12 h`), an LP-layer dual of `1.20` on the node power-balance constraint reads directly as `$1.20/kWh` at that node and time. The equivalent power-unit dual would have been `1.20 * (1/12) = 0.10 $/kW`, but HAEO no longer surfaces this intermediate form.

## Test plan

- [x] `uv run pytest custom_components/haeo` – 1613 passed, 2 skipped
- [x] `uv run pytest tests/scenarios` – 64 passed (6 snapshots unchanged)
- [x] `uv run ruff check` – clean
- [x] `uv run ruff format --check` – clean
- [x] `uv run pyright` – 0 errors, 0 warnings
- [x] `uv run lint-imports` – 4 contracts kept
- [x] `uv run mdformat --check docs/modeling/shadow-prices.md` – clean

## Credit

The LP-layer reformulation is @TrentHouliston's work, cherry-picked from `experiment/energy-native-constraints` (commit `af5ccb9`) with author attribution preserved. This PR's follow-up commit is the adapter / tests / docs cleanup that the new formulation enables.


purcell-lab:
Running this PR locally without issue — happy to take feedback on naming, scope, or whether to fold the battery case in here vs a follow-up.

Pre-release available as a HACS custom repository for testing: add https://github.com/purcell-lab/haeo, then download tag v0.4.0rc6.

Note: the failing Tests check is the Codecov tokenless-upload step (1733 tests pass); fork PRs to hass-energy/haeo can't currently complete that upload. Happy to adjust the workflow if helpful.

TrentHouliston:
I think we should just output the time normalised value, it's far more useful. I'm not sure I'd ever use the power based one, as since haeo has different time slices, you can't compare the value between horizon timepoints. I did want to move them into diagnostics too, so that part is good

TrentHouliston:
Actually after a little experimenting, I think I should just be representing my constraints in energy units natively. Then the values come back in $/kWh naturally.

purcell-lab:
@TrentHouliston — applied your feedback. Force-pushed `a77d77d` to this branch:

- **Dropped every `$/kW` shadow sensor** across all adapters (battery, battery_section, grid, inverter, load, node, solar). Verified `grep` over `core/adapters/elements/*.py` returns zero `$/kW` publications.
- **Renamed** all retained `$/kWh` sensors to the `*_shadow_energy_price` suffix for grammatical clarity.
- **Dropped** the unused `shadow_price_per_power()` helper.
- LP model layer untouched — raw duals continue to compute in `$/kW`, the adapter layer converts via `shadow_price_per_energy()` before publishing.

On your "switch to energy-balance constraints" idea — flagged that as the follow-up in both the commit body and PR description. Happy to take that on as a separate PR once this lands; the user-facing sensor surface won't change, only the internals.

All checks green (1623 + 162 tests pass, ruff/pyright/lint-imports/mdformat clean). Ready for re-review.

purcell-lab:
Force-pushed a refactor that adopts @TrentHouliston's energy-native LP approach from [`experiment/energy-native-constraints`](https://github.com/hass-energy/haeo/compare/main...experiment/energy-native-constraints). Trent suggested the switch from power to energy units could be very simple at the LP layer, and after digging in I agreed — it's a much cleaner fix than the adapter-layer division this PR originally proposed.

**What changed at the LP layer (Trent's commit, cherry-picked with author preserved):**

- `element.py` and `elements/segments/power_limit.py` now multiply their constraints by `dt` before handing them to HiGHS.
- Result: the dual variables come out of the solver in `$/kWh` natively, with no post-processing needed.

**What I removed on top of that:**

- The `shadow_price_utils.py` helper module + its tests.
- `shadow_price_per_energy` calls and `numpy/NDArray` imports from every adapter.
- The explicit `periods` parameter from adapter `outputs()` signatures that no longer need it.

Net: −311 / +72 lines, and the "battery deferred" caveat from the original PR is now resolved because `BATTERY_POWER_BALANCE := ELEMENT_POWER_BALANCE` inherits the dt-scaling for free.

QA:

- `uv run pytest custom_components/haeo` – 1613 passed, 2 skipped
- `uv run pytest tests/scenarios` – 64 passed, 6 snapshots unchanged
- `ruff`, `pyright`, `lint-imports`, `mdformat` – all clean

The primal solution is unchanged (multiplying both sides of every constraint by `dt > 0` is a strict scaling), so existing scenario snapshots pass without an update. Thanks @TrentHouliston for the elegant fix.

PR #425, opened 2026-05-09
feat(load): add forecast statistics sensors

purcell-lab:
## Summary

Adds 8 forecast statistics sensors per Load element so users can compare optimiser dispatch against alternative load schedules, and plumbs an optional `display_precision` through `CoordinatorOutput` so dollar-denominated sensors render with two decimal places.

> **Refactored 2026-06-01:** the original threshold-price work has been **dropped from this PR** now that #441 has landed the `InputStore` refactor and `consumption_cost` `SurfacedPriceHint` on Load — those provide a cleaner replacement for the threshold mechanism. This PR is now scoped to load statistics sensors only.

## Sensors

Four pairs of cumulative + rolling-24h stats are published per Load:

| Sensor | Unit | Notes |
|---|---|---|
| `load_horizon_energy` | kWh | Cumulative energy series; state = horizon total |
| `load_horizon_marginal_cost` | $ | display_precision=2 |
| `load_horizon_runtime` | h | Timesteps with dispatched power > ε |
| `load_horizon_average_marginal_price` | $/kWh | Energy-weighted; 0 when no energy dispatched |
| `load_next_24h_energy` | kWh | Rolling 24h from horizon origin (scalar broadcast) |
| `load_next_24h_marginal_cost` | $ | display_precision=2 |
| `load_next_24h_runtime` | h | |
| `load_next_24h_average_marginal_price` | $/kWh | display_precision=2 |

Pre-existing `load_power` and `load_forecast_limit_price` are unchanged.

## How marginal cost is computed

For each timestep the dispatched power is integrated against the **source-node `element_power_balance` dual** (already period-integrated by the LP objective, $/kWh):

```
cost_t = power_t * node_dual_t        # units: $ per timestep
runtime_t = 1 if power_t > ε else 0
avg_price = sum(cost) / sum(energy)   # 0 if sum(energy) == 0
```

If the source node has no dual (e.g. an unconstrained pass-through topology) the stats sensors are **omitted entirely** rather than publishing a misleading zero-cost figure.

## Adapter contract

`LoadAdapter.outputs()` gains an optional `periods: Sequence[float] = ()` keyword argument used to integrate `power * dual` over time. No change is required for adapters that do not consume periods.

## Display precision plumbing

- New optional `OutputData.display_precision: int | None = None`
- New optional `CoordinatorOutput.display_precision: int | None = None`, passed through `_build_coordinator_output`
- `HaeoSensor` sets `self._attr_suggested_display_precision = output.display_precision` when provided
- The four dollar-denominated sensors above set `display_precision=2`

## Test coverage

- 9 unit tests in `custom_components/haeo/core/adapters/elements/tests/test_load_stats.py` (zero-energy, all-zero-dual, fixed-load, ε-runtime, missing-dual omission, etc.)
- Scenario snapshots refreshed across all 6 scenarios

## Out of scope (intentionally dropped from this PR)

- Threshold-price `SurfacedPriceHint` and config-flow section — superseded by `consumption_cost` from #441
- `shedding.md` doc and threshold LP integration tests — removed with the threshold mechanism
- Schema migration for the threshold field — no longer needed

## Related

- #441 — InputStore extraction + `consumption_cost` SurfacedPriceHint (the replacement for our threshold mechanism)
- #430 — surfaced policy pricing fields on battery and load config flows

## Test plan

- [x] `uv run pytest custom_components/haeo/ tests/ --ignore=tests/guides` — 1847 passed, 2 skipped
- [x] `uv run ruff check custom_components/ tests/` — clean
- [x] `uv run ruff format --check custom_components/ tests/` — clean
- [x] `uv run pyright` — 0 errors
- [x] Scenario snapshots regenerated


purcell-lab:
image

threshold_validation

purcell-lab:
Force-push: fix for stats sensors silently dropping on source+sink buses
Found a real bug in this PR while smoke-testing locally on a live HA install (0.4.0rc5.post10 build): all 8 cumulative + rolling-24h stats sensors (load_total_*, load_daily_*) were registering correctly in tests but never reaching the HA entity registry in production.

Root cause
After #426 (DEFAULT_TAG removal), element_power_balance returns a multi-block per-tag dual vector — for a switchboard node with is_source=is_sink=True (the HA default) on a 4-period horizon, the length is 12, not 4. The stats helper checked if len(values) != n_periods: return None, silently dropping all stats outputs whenever the load was attached to anything other than a pure junction. Threshold price kept working, which is why the bug was easy to miss — it looked almost-right.

Fix
Sum across the dual blocks to recover a single $/kWh-per-timestep series. Production/consumption decomposition duals are 0 at the optimum, so the summed value matches the prior single-tag behaviour where this loop only saw one block. Added a regression test exercising bus_is_source=bus_is_sink=True to lock the behaviour in.

What's in the force-push
Squashed into feat(load): add cumulative + rolling-24h stats sensors (the dual-shape handling): the _node_dual_values change + the new regression test

Squashed into test(scenarios): refresh outputs snapshots for new load stats sensors: scenario6 snapshot refresh (it uses a real switchboard topology, so the new sensors now appear in its snapshot)

Stack remains 8 commits, no merge conflicts, fully rebased on current main.

Follow-up (separate issue incoming)
The inverter adapter has the same underlying bug — inverter.py:145 reads dc_bus[ELEMENT_POWER_BALANCE] directly with no shape handling. It doesn't fail visibly because the inverter just passes whatever shape through, but the resulting DC-bus shadow price sensor is wrong for multi-tag topologies. Filing as a separate issue rather than expanding this PR's scope.

purcell-lab:
## Update: stats sensors refactor + display precision (post11 → post14)

Force-pushed a stack of improvements on top of the earlier multi-block-dual fix. The 8-commit stack is unchanged in shape — these all squashed into the existing **feat(load): add cumulative + rolling-24h stats sensors** commit. Head is now `ddb76642`.

**READY FOR MERGE**

Summary of what's new since the last review pass:

### 1. Cost over-statement fix (validated on a real install)

For Switchboard-style passthrough nodes, the LP emits a multi-block `element_power_balance` dual — one block per VLAN tag — concatenated into a single vector. The previous code summed all `K` blocks, over-stating the load cost by a factor of `K`. The fix slices `values[:n_periods]`, matching what the coordinator publishes on the corresponding HA sensor (which zips with `n_periods+1` timestamps using `strict=False`, dropping trailing tag-balance blocks).

On a live install with a 3-VLAN Switchboard this surfaced as a ~3.3× over-statement on the horizon cost; post-fix the sensor reads `$5.90` against ~58 kWh served (≈ $0.10/kWh average), which matches both the dispatched profile and the source-node dual.

A regression test (`test_stats_outputs_present_when_bus_is_source_and_sink`) solves a source+sink multi-block-dual case and asserts the 8 sensors still surface with the correct per-interval values.

### 2. Sensor renames — breaking change

After live-install feedback that "cost" was ambiguous when the value was a $/kWh average (i.e. a marginal price, not a cost), the four monetary stats sensors were renamed:

| Old                              | New                                       |
| -------------------------------- | ----------------------------------------- |
| `load_horizon_cost`              | `load_horizon_marginal_cost`              |
| `load_horizon_average_cost`      | `load_horizon_average_marginal_price`     |
| `load_next_24h_cost`             | `load_next_24h_marginal_cost`             |
| `load_next_24h_average_cost`     | `load_next_24h_average_marginal_price`    |

The dollar totals keep `cost` in the name (unit `$`). The $/kWh averages now read `marginal_price`, matching the underlying shadow-price quantity. `*_energy`, `*_runtime`, `load_power`, `load_forecast_limit_price`, and `load_threshold_price` are unchanged.

Translations (`en.json`) and the 6 scenario `outputs.json` snapshots have been refreshed accordingly.

### 3. Per-interval forecast values + dedicated state override

The previous draft published either a cumulative running total (`*_cost`/`*_energy`/`*_runtime`) or a scalar broadcast across every step (`next_24h_*`) in the sensor's `forecast` attribute, which made ApexCharts trend cards render as a stepped staircase or a flat repeating line. The new shape:

- `forecast` attribute → **per-interval** contribution to the window
- entity `state` → integrated total across the horizon (or next-24h window)

This is implemented via a new optional `OutputData.state` scalar override that takes precedence over `state_last` / `values[0]`. The coordinator (`_build_coordinator_output`) honours it, so the forecast time-series and the state value can describe different views of the same series without overloading `values`.

Per-sensor semantics:
- `*_marginal_cost` forecast: cost incurred in each interval ($); state = sum across window
- `*_energy` forecast: energy served in each interval (kWh); state = sum across window
- `*_runtime` forecast: runtime allocated in each interval (h); state = sum across window
- `*_average_marginal_price` forecast: per-interval shadow price ($/kWh); state = energy-weighted average

### 4. Suggested display precision (new, optional)

Added an optional `display_precision: int | None` field to `OutputData`, flowing through `CoordinatorOutput` onto `SensorEntity._attr_suggested_display_precision`. The four dollar-denominated load stats sensors now suggest **2 decimal places** out of the box (`$5.90`, `$0.10/kWh`), rather than relying on the integration's smart-rounding default (~4 sig figs, which can drift into 3- or 4-decimal precision on small $/kWh averages).

`display_precision=None` (default) preserves the existing smart-rounding behaviour for energy/runtime/power sensors — no other sensors are affected. Users can still override per-entity in the Home Assistant UI.

### Verification

- `uv run pytest --ignore=tests/guides` — **1865 passed, 2 skipped** (includes refreshed scenario snapshots).
- `uv run ruff check . && uv run ruff format --check .` — clean.
- `uv run pyright` — 0 errors.
- Live install soak: post11 → post12 → post13 → post14 builds verified. Sensors render with the new names, totals/averages are stable and on the right order of magnitude, ApexCharts trend cards now render true per-interval cost/energy profiles, and the four dollar sensors pin to 2 dp without per-entity overrides.

### Backward-compatibility note for users

Anyone with dashboards or automations referencing `sensor.<load>_horizon_cost`, `sensor.<load>_horizon_average_cost`, `sensor.<load>_next_24h_cost`, or `sensor.<load>_next_24h_average_cost` will need to update those references on upgrade. Energy/runtime/power/threshold entity_ids are unchanged.

I'll flag this in the release notes when this branch ships in a tagged release.

Screenshot 2026-05-19 14 13 36

purcell-lab:
I think I have doubled up on some of the scope of https://github.com/hass-energy/haeo/pull/430 with my threshold sensor.

They appear to do the same thing.

A threshold of say 5c/kWh when I want to schedule my pool heater seems to have the same effect as a cost of -5c/ kWh.

I still this this PR has some merit in surfacing the additional stats sensors, but the threshold price sensor is a double and could be removed.  Although specifying a negative cost might be confusing.  Similarly for my crypto miner, I have set the threshold to the price sensor, I would have to set the cost to the negative price sensor (which would need another template).

TrentHouliston:
There was one more change I wanted to do that might influence this change. Then I'll look at how the best way to surface the new outputs are. I agree that some of these are quite useful

purcell-lab:
## Refactored: rebased onto `main` (now `0.4.0rc6`), scope narrowed to load statistics sensors only

Force-pushed `feat/load-threshold-price` → now `0cad38ca`. Branch name is retained but no longer reflects the contents — happy to rename if preferred.

### What changed in this refactor

**Dropped** (now redundant with #441):
- Threshold-price `SurfacedPriceHint` on Load
- Config-flow threshold section
- `docs/modeling/shedding.md`
- Threshold LP integration test file
- Schema migration for the threshold field

The `InputStore` + `consumption_cost` `SurfacedPriceHint` you landed in #441 provides a cleaner replacement for what the threshold was doing — a single backing store where setting a value on a Load mirrors between policy and entity, with the price hint flowing through the existing policy pricing path.

**Kept and rebased cleanly onto current `main`:**
- 8 forecast statistics sensors per Load (cumulative horizon + rolling 24h, each emitting energy / marginal_cost / runtime / average_marginal_price)
- `OutputData.display_precision` + `CoordinatorOutput.display_precision` plumbing
- `display_precision=2` on the four dollar-denominated sensors
- `LoadAdapter.outputs(..., periods=…)` integration of `power * source_node_dual` for marginal cost
- 9 unit tests in `test_load_stats.py`
- Scenario snapshots refreshed across all 6 scenarios (large additions are just the 8 new sensors × periods)

### Commits

```
0cad38ca test(scenarios): refresh outputs snapshots for new load stats sensors
01917510 feat(load): add forecast statistics sensors
```

Both carry `Co-authored-by: TrentHouliston <trent@houliston.me>`.

### Validation

- `uv run pytest custom_components/haeo/ tests/ --ignore=tests/guides` → **1847 passed, 2 skipped**
- `uv run ruff check custom_components/ tests/` → clean
- `uv run ruff format --check custom_components/ tests/` → clean
- `uv run pyright` → **0 errors**

Ready for another look when you have a moment.

PR #433, opened 2026-05-15
Grid operator signals: marginal cost, capacity, and intent engine

TrentHouliston:
## Summary

Adds grid operator signal outputs derived from LP reduced costs and column/row ranging, plus a new intent engine for connection control recommendations.

### Intent engine (`core/model/intent.py`)
New module that uses HiGHS column ranging on connection `power_in` variables to derive per-period control recommendations:
- **UNLIMIT**: hardware can run free (range reaches forecast max with non-positive reduced cost)
- **SET**: hardware should be actively limited to the recommended value
- **UNKNOWN**: zero forecast — model has no information, adapter decides based on context

Each recommendation includes the optimal value, limit, safe operating band `[band_min, band_max]`, reduced cost ($/kWh), and forecast max.

### Connection outputs (`core/model/elements/connection.py`)
Two new outputs on every connection:
- **`connection_marginal_cost`**: reduced cost of `power_in` variables converted to $/kWh, aggregated across tags (max). Returns `None` when all values are near-zero.
- **`connection_capacity`**: available capacity before a basis change (price step), derived from column ranging. Reported in kWh.

### Constraint ranging (`core/model/reactive/decorators.py`)
Shadow price outputs from reactive constraints now include `range_up` and `range_dn` fields on `OutputData`, indicating how much the constraint RHS can change before the shadow price changes.

### Grid adapter (`core/adapters/elements/grid.py`)
Four new grid-level outputs mapped from the connection outputs:
- `grid_forced_import_cost` / `grid_forced_export_cost` — marginal cost signals
- `grid_import_headroom` / `grid_export_headroom` — capacity signals

### Robustness fixes
- `entity_value.py`: Accept `Sequence[str]` (not just `list[str]`) since HA's config entry deep-freeze converts lists to tuples
- `elements/__init__.py`: Accept `Sequence` where `list` was expected, for the same reason
- `config_loader.py`: Use `Mapping` instead of `dict` for isinstance checks; `Sequence[str]` for entity values
- HiGHS type stubs: Added `getRanging()`, `HighsRanging`, `HighsRangingRecord`

### Tests
- 11 new tests covering intent engine scenarios (solar unlimit/set, zero forecast, battery, grid connections)

### Documentation
- Intent engine developer guide
- Updated connection modeling docs
- Grid element user guide additions
- Historic diagnostics walkthrough


purcell-lab:
Nice work on `compute_intent`. The ranging plumbing is clean, the `UNLIMIT/SET/UNKNOWN` decision rule reads well, and the zero-forecast disambiguation via `constraint_shadow` is the right call. One gap I'd like to flag for a follow-up PR, motivated by an upcoming regulatory change rather than anything in this branch.

### The gap: single-band output vs multi-band offer requirements

`ConnectionIntent` exposes one `(band_min, band_max)` pair per period at the LP optimum cost. That is exactly the right shape for a hardware control signal ("here is the safe operating band at today's price"), and it is what the engine should send to the adapter on every cycle.

It is not yet the right shape for a wholesale **offer curve**, which needs multiple `(price, kW)` pairs spanning the price space the market clears against. With one band per period the curve has one step in it: above `cost_up` we go to zero (or the lower bound), below `cost_dn` we go to forecast max. Everything in between is implied to be at the LP optimum. For IPRR Stage 2 upward offers (price ceiling at MPC, currently $20/kWh per [AEMC 2026-27](https://www.aemc.gov.au/news-centre/media-releases/aemc-updates-market-price-cap-2026-27)) and for the Reliability Panel's April 2026 MSL3 reform (price floor pinned to MFP at -$1,000/MWh, [Reliability Panel rule change](https://www.aemc.gov.au/sites/default/files/2026-04/new_rule_change_proposal_-_reliability_panel_-_minimum_system_load-_20260423.pdf)) the offer needs intermediate steps. Load that comes on at -$1.00/kWh is a different commitment from load that comes on at -$0.10/kWh, and an export that fires at $1/kWh is a different commitment from one that fires at $5/kWh.

### Proposed follow-up: parametric sweep on top of this PR

The good news is the LP and the basis are already there. A parametric sweep of `solver.changeColCost` across a price grid, warm-starting from the saved basis on each resolve, gives a staircase offer curve at marginal cost, typically under 50 ms per period for the household-scale models we run. Sketch:

```python
basis = solver.getBasis()
curve = []
for p in price_grid:  # e.g. [MFP, -0.5, -0.1, 0, retail, 1.0, 5.0, MPC]
    solver.changeColCost(power_in[t].index, p * dt[t])
    solver.setBasis(basis)
    solver.run()
    sol = solver.getSolution()
    curve.append((p, sol.col_value[power_in[t].index]))
```

I have a working sketch on top of this branch (separate WIP) that adds a `connection_offer_curve` output alongside the existing `marginal_cost` and `capacity` outputs, with tests covering the symmetric load-up (MFP-anchored) and shed-up (MPC-anchored) cases. Happy to open it as a follow-up PR once this one lands. Two open questions for you before I do:

1. Do you want the offer curve as a first-class output on `ConnectionElement`, or as a separate `OfferCurve` element that consumes the same solver state? The first matches `marginal_cost_output`'s layout, the second decouples it cleanly if downstream code wants to evaluate offer curves for elements that are not connections (storage, flexible loads).
2. Is the price grid configurable per-instance, or should we ship a sensible default (something like `[MFP, retail Ã— {-3, -1, 0, 1, 3}, MPC]`)? My preference is configurable with a default, but happy to defer.

Either way, nothing in this PR needs to change for the follow-up to fit. Calling it out here so the design intent is on the record.

Mark

PR #468, opened 2026-06-05
fix(#467): propagate live subentry value edits into LP via TrackedParam updater chain

### LINKED ISSUE #467: Battery min SOC (and other battery limit/SOC-pricing) settings do not take effect without integration reload
## Summary

Changing `number.battery_min_charge` (and other battery limit/SOC-pricing entities) does not propagate to the optimizer's effective bound without a full integration reload. The entity reflects the user's input correctly, but `OptimizationContext.participants.Battery.limits.min_charge_percentage` remains stuck on the load-time value. This is the same class of bug PR #408 fixed for grid policy pricing.

## Environment

- HAEO `main` (post-merge of #425, commit `4c8ab869`)
- Home Assistant Core (current stable)
- Battery subentry configured with `min_charge_percentage` exposed as a constant value (number entity)

## Steps to reproduce

1. Configure a hub with a Battery subentry.
2. Set `number.battery_min_charge` to a value (e.g. `50`). Let the integration optimize and settle.
3. Capture a diagnostics dump (`action: haeo.save_diagnostics`) — note `OptimizationContext.participants.Battery.limits.min_charge_percentage` matches the entity value.
4. Change `number.battery_min_charge` to a different value (e.g. `20`).
5. Wait past the debounce; let the next optimization cycle run.
6. Capture another diagnostics dump.

## Expected

`OptimizationContext.participants.Battery.limits.min_charge_percentage` reflects the new value (e.g. `0.2` after step 4), and the optimizer schedules against the new floor.

## Actual

`OptimizationContext.participants.Battery.limits.min_charge_percentage` remains at the load-time value across all subsequent dumps. Only an integration reload makes the new value take effect.

Observed across five consecutive diagnostic dumps from one live instance:

| # | `number.battery_min_charge` | `participants.Battery.limits.min_charge_percentage` | `battery_soc_min_shadow_price` |
|---|---|---|---|
| 1 | 0.0 | 0.0 | 0.0 |
| 2 | 20.0 | **0.0** (stale) | 0.0 |
| 3 | 35.0 | **50.0** (stale) | 0.0 |
| 4 | 35.0 | **50.0** (stale) | 0.0 |
| 5 | 20.0 | **50.0** (stale) | 0.0 |

No `haeo` errors are logged. The subentry update is accepted and persisted to `subentry.data` (via `util.async_update_subentry_value`), but the running coordinator's network is not updated in-flight.

### Reactive plumbing is correct in isolation

Two unit tests were added (`coordinator/tests/test_battery_min_charge_repro.py`):

- `test_battery_updater_recomputes_capacity_on_min_charge_change` — **PASSES**. Confirms `_build_element_updater` correctly runs the battery adapter, recomputes `capacity_range = (max - min) * capacity_first` and `initial_charge = (initial_soc - min) * capacity_first`, and writes them through to the live `Battery` model element's `TrackedParam` descriptors.
- `test_store_change_queues_pending_update_with_correct_min_charge` — **PASSES**. Confirms that `store.set_value(20.0)` synchronously fires `_notify()`, the coordinator's `_handle_element_update("Battery")` listener runs, `_load_element_config("Battery")` reads the new live value via `_field_values_for_element`, and the resulting config in `_pending_element_updates["Battery"]` correctly contains `min_charge_percentage = [0.2, 0.2, 0.2]`. `signal_optimization_stale` is called exactly once.

So both halves of the reactive chain work correctly in isolation.

### Snapshot is not the leak

`coordinator/coordinator.py:318-321` builds `_participant_configs` once at `__init__`, but `_load_element_config` (L628-657) merges this snapshot with live values from `_field_values_for_element` (L617-626) before queueing. The snapshot itself is rebuilt with live values on every listener fire.

### Leading hypotheses

1. **`_optimization_in_progress` early-return race** (`coordinator.py:711-717`) — `_async_update_data` returns early when an optimization is already in flight. If a `signal_optimization_stale()` triggered by a store change hits this early-return, the pending dict can be drained at the wrong time, causing the network to remain on stale bounds while the queue is cleared.

2. **Listener key mismatch** — the listener is keyed by `subentry.title`. If a user's battery subentry title differs from the element name used in `_pending_element_updates`, the queue would never receive the update.

### Related but separate bug: SOC pricing fields cannot be updated reactively at all

`core/model/elements/segments/soc_pricing.py:55-58` — `_discharge_energy_threshold`, `_charge_capacity_threshold`, `_discharge_energy_price`, `_charge_capacity_price` are plain instance attributes, not `TrackedParam`. This means the following battery fields require a full integration reload regardless of any min_charge fix:

- `battery.undercharge.percentage`
- `battery.undercharge.cost`
- `battery.overcharge.percentage`
- `battery.overcharge.cost`

A broader audit identified that `battery.limits.max_charge_percentage` shares the same `TrackedParam` path as `min_charge_percentage` and is affected by the same coordinator-level race.

## Suggested fix direction

Mirror the PR #408 pattern. Two parts:

**Part A — Battery limits (TrackedParam path).** Ensure pending updates are applied even when an optimization is in flight, e.g. by short-circuiting the early-return only when `_pending_element_updates` is empty:

```python
async def _async_update_data(self):
    if self._optimization_in_progress and not self._pending_element_updates:
        return self._last_data
    self._apply_pending_element_updates()
    # ...
```

**Part B — SOC pricing (separate PR).** Promote the four `soc_pricing` thresholds/prices to `TrackedParam` and extend the battery adapter to write fresh undercharge/overcharge values through the updater the same way capacity is handled.

Happy to open a draft PR for Part A once the integrated unit test (`test_apply_pending_updates_writes_to_network`) is finalized to lock the fix in place.

purcell-lab:
## Summary

Fixes #467. Editing `number.battery_min_charge` (and other battery limit / SOC-pricing entities) now propagates to the live LP without an integration reload.

The bug had two distinct root causes — fixing only one of them left the other still gating the live update. Both are addressed here.

## Root cause #1 — value-edit listener never refreshed TrackedParams

Editable `number` / `switch` entities persist their new value via `util.async_update_subentry_value`, which mutates the subentry's data and triggers `async_update_listener`, which calls `signal_optimization_stale`. But nothing on that chain populated `coordinator._pending_element_updates`, so `_apply_pending_element_updates` ran as a no-op and the network's `TrackedParam`s kept their old values across optimization runs.

### Fix
- `async_update_subentry_value` now also records the changed `subentry_id` on `runtime_data.value_update_subentry_id`.
- `async_update_listener` reads that id when handling a value-only update and asks the coordinator to refresh the affected participant via the new `handle_subentry_value_update()` method before signalling optimization stale.
- `handle_subentry_value_update()` reverse-looks-up the element name from the `subentry_id` (or refreshes all participants when the id is `None`) and re-loads each element config into `_pending_element_updates`, so the next optimization picks up the new values via the existing `_apply_pending_element_updates` → `updater(element_config)` → `TrackedParam` write chain.

## Root cause #2 — `SocPricingSegment` params were not `TrackedParam`

After commit 1, the listener correctly fired and the updater chain ran — but the LP still didn't respect a new `min_charge` floor. Reason: `SocPricingSegment` stored `discharge_energy_threshold` / `charge_capacity_threshold` / `discharge_energy_price` / `charge_capacity_price` as plain instance attributes (`_discharge_energy_threshold` etc.). `_discover_setters` in `coordinator/network.py` only captures **class-level `TrackedParam` descriptors** (via `getattr(type(obj), key, None)`), so the captured updater chain wrote *nothing* for `soc_pricing`.

In the affected configuration, `min_charge_percentage` only reaches the LP via the `soc_pricing` segment's `discharge_energy_threshold` (when `undercharge.percentage = 0`, the existing `lower_ratio` path is dormant), so an untracked threshold field meant the floor was effectively pinned at integration load time.

### Fix
- Promote `discharge_energy_threshold`, `charge_capacity_threshold`, `discharge_energy_price`, `charge_capacity_price` to class-level `TrackedParam` descriptors on `SocPricingSegment`.
- Always create the discharge/charge slack LP variables so the model structure stays stable across runs even when a price transitions between `None` and a numeric value.
- Gate constraint emission on `(threshold AND price)` rather than slack-variable presence so partial configurations no longer pin SOC unintentionally.
- Preserve the historical guard: a price set without a corresponding threshold still raises `ValueError` at init.

### Unit tests

All `custom_components/haeo` tests pass: **1751 passed, 2 skipped** (was 1749 pre-fix). Added:

- `test_handle_subentry_value_update_scopes_to_matching_participant`
- `test_handle_subentry_value_update_with_none_refreshes_all_participants`
- `test_handle_subentry_value_update_ignores_unknown_subentry`
- `test_discover_setters_finds_soc_pricing_thresholds`
- `test_soc_pricing_setters_write_fresh_threshold`
- Extended `test_flag_remains_set_after_call` to also assert `subentry_id` is recorded; `test_flag_cleared_on_exception` asserts it is cleared on failure paths.

### Live reproduction on a real instance

Repro on `0.4.0rc7` baseline (pre-fix):

| Step | tail_min SoC | Verdict |
|---|---|---|
| Phase A — set `min_charge` to 10/40/25 without reload | 0.00% | FAIL (3/3) |
| Post-reload at 25 | 24.62% | PASS |
| Phase B — set `min_charge` to 40 after reload | 25.00% (stuck) | FAIL |

Re-run on a build with both commits applied (`0.4.0rc7.post3`):

| Step | requested | tail_min SoC | verdict |
|---|---|---|---|
| Phase A set→10 (no reload) | 10 | 9.65% | PASS |
| Phase A set→40 (no reload) | 40 | 40.00% | PASS |
| Phase A set→25 (no reload) | 25 | 25.00% | PASS |
| Phase B set→10 (post reload) | 10 | 10.00% | PASS |
| Phase B set→40 (post reload) | 40 | 40.00% | PASS |
| Phase B set→25 (post reload) | 25 | 25.00% | PASS |

All 10 phase steps PASS. Repro script output: `NO ISSUE: Both phases passed. Fix is fully effective.`

## Commits

1. `fix(coordinator): push subentry value edits into network TrackedParams` — fixes root cause #1 (the listener chain).
2. `fix(model): make SocPricingSegment params TrackedParam for live updates` — fixes root cause #2 (the segment-level descriptor gap).

## Notes for reviewers

- No public API breaks. `_discover_setters` already understands `TrackedParam` descriptors, so the SOC-pricing setters slot in transparently.
- Constraint behaviour for prior-valid configs is unchanged: when `(threshold, price)` are both set, the same slack/cost constraints emit; when both are absent, no constraints emit (same as before); when only one is set, init still raises `ValueError`.
- The slack-variable creation is now unconditional. The cost contribution and the slack-bound constraint remain gated on `(threshold AND price)`, so an absent price/threshold still contributes nothing.
- `discharge_energy_slack` and `charge_capacity_slack` public properties still return `None` when their respective `(price, threshold)` pair is absent, preserving the read-side contract for existing callers.

Closes #467

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -226,7 +226,7 @@ jobs:
 
       - name: Upload frontend coverage to Codecov
         if: ${{ github.event_name != 'release' && !cancelled() }}
-        uses: codecov/codecov-action@671740ac38dd9b0130fbe1cec585b89eea48d3de # v5.5.2
+        uses: codecov/codecov-action@e53489f4d376d79066609109e7a95a29eb3740b1 # v7.0.0
         with:
           files: frontend/haeo-forecast-card/coverage/lcov.info
           flags: frontend
@@ -256,13 +256,13 @@ jobs:
 
       - name: Upload test results to Codecov
         if: ${{ github.event_name != 'release' && !cancelled() }}
-        uses: codecov/codecov-action@671740ac38dd9b0130fbe1cec585b89eea48d3de # v5.5.2
+        uses: codecov/codecov-action@e53489f4d376d79066609109e7a95a29eb3740b1 # v7.0.0
         with:
           report_type: test_results
 
       - name: Upload coverage to Codecov
         if: ${{ github.event_name != 'release' && !cancelled() }}
-        uses: codecov/codecov-action@671740ac38dd9b0130fbe1cec585b89eea48d3de # v5.5.2
+        uses: codecov/codecov-action@e53489f4d376d79066609109e7a95a29eb3740b1 # v7.0.0
         with:
           flags: python
           fail_ci_if_error: true
--- a/custom_components/haeo/coordinator/tests/test_network.py
+++ b/custom_components/haeo/coordinator/tests/test_network.py
@@ -28,6 +28,7 @@ from custom_components.haeo.core.model.elements.policy_pricing import (
     PolicyPricingTerm,
 )
 from custom_components.haeo.core.model.elements.segments import EfficiencySegment, PowerLimitSegment
+from custom_components.haeo.core.model.elements.segments.soc_pricing import SocPricingSegment
 from custom_components.haeo.core.schema import as_connection_target
 from custom_components.haeo.core.schema.elements import ElementConfigData, ElementType
 from custom_components.haeo.core.schema.elements.connection import (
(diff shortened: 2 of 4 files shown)


TrentHouliston:
Maintainer cleanup pushed to this branch (force-push):

- **Dropped commit 1** (`fix(coordinator): push subentry value edits into network TrackedParams`). After red/green testing, that path turned out to be redundant: the existing `InputStore` listener (`_subscribe_to_input_stores` → `_handle_element_update`) already enqueues `_pending_element_updates` on a value edit, so the extra plumbing wasn't doing anything new.
- **Kept commit 2** (`make SocPricingSegment params TrackedParam for live updates`) unchanged — that is the actual fix for #467. The battery `min_charge` floor reaches the LP through the discharge `soc_pricing` threshold, which wasn't a `TrackedParam`, so edits didn't propagate without a reload.
- **Added a simplification commit**: merged the two per-direction slack-bound constraints into a single `soc_slack_bounds` and factored the slack-exposure guard into a small helper (behavior unchanged). Kept the explicit per-field validation rather than a dynamic loop for readability.
- **Added a regression test** `test_soc_pricing_live_update.py`: edits a battery `min_charge` and asserts the live discharge `soc_pricing` threshold updates through the element updater. Verified red on the pre-fix baseline (`AttributeError` — no `discharge_energy_threshold` TrackedParam) and green with commit 2.

Separately, the "battery created with no device/entities" issue is unrelated to #467 and is fixed in #472.

PR #476, opened 2026-06-20
chore(quality-scale): correct self-assessment and align platforms

purcell-lab:
A small housekeeping pass on `custom_components/haeo/quality_scale.yaml` plus two tiny platform edits.

While doing an end-to-end review of the integration against the [Home Assistant Quality Scale](https://developers.home-assistant.io/docs/core/integration-quality-scale/) (Bronze through Platinum), I noticed a handful of rows where the self-assessment was more conservative than the actual state of the code. None of these changes are behavioural — they just bring the file into line with what's already shipping.

### `custom_components/haeo/quality_scale.yaml`

| Rule | Before | After | Why |
|---|---|---|---|
| `action-setup` (Bronze) | `not-applicable` "Integration exposes no Home Assistant services" | `done` | `services.py` registers `haeo.save_diagnostics` and `haeo.optimize`; both are declared in `services.yaml` with `config_entry` selectors and translated under `services` in `translations/en.json`. |
| `action-exceptions` (Silver) | `not-applicable` "Integration registers no service handlers" | `done` | Both handlers raise `ServiceValidationError` with `translation_domain="haeo"`, `translation_key`, and `translation_placeholders`; the keys live under `exceptions` in `translations/en.json`. |
| `brands` (Bronze) | `external` "Branding assets must be contributed to the home-assistant/brands repository outside this project" | `done` | The icon and logo are already published in `home-assistant/brands` at `custom_integrations/haeo/` — both return HTTP 200. |
| `docs-supported-functions` (Gold) | comment referenced `docs/reference/sensors.md` (doesn't exist) | comment rewritten to point at `docs/modeling/device-layer/` and `docs/user-guide/elements/` | Those pages already enumerate every published output per element. |
| `docs-known-limitations` (Gold) | comment referenced `docs/reference/limitations.md` (doesn't exist) | comment rewritten to point at `docs/user-guide/alternatives.md` and `docs/modeling/` | Those pages already capture HAEO's scope boundaries and modelling assumptions. |
| File path references throughout | older `coordinator.py`, `sensors/__init__.py`, `data/__init__.py` layout | `coordinator/coordinator.py`, `entities/haeo_sensor.py`, `diagnostics/__init__.py`, `flows/options.py`, `entities/device.py` | The integration was reorganised but the assessment hadn't caught up. |
| `target_tier` | `platinum` | `gold` | Platinum's own four rules are already met (strict pyright, executor offload for the HiGHS solve, no HTTP), but Platinum requires Gold — so Gold is the realistic next milestone. |
| `summary` / `blockers` / `notes` | total 51, done 37, exempt 7, not-applicable 2, external 1, todo 4 | total 51, done 40, exempt 7, not-applicable 0, external 0, todo 1 | Reflects the corrections above. |

The only Gold rule still listed as `todo` is `entity-disabled-by-default` for diagnostic auxiliaries — that's a design call left for separate review.

### `manifest.json`

Added `"quality_scale": "silver"`. This is the canonical signal for HA core's integrations panel to surface the tier badge. It matches the verified state of `quality_scale.yaml`.

### `number.py`, `switch.py`

Added `PARALLEL_UPDATES = 0` matching `sensor.py`, with a comment explaining why (InputStore-backed, no external I/O). This makes the silver `parallel-updates` rule trivially compliant across every platform the integration registers, instead of just `sensor`.

### Test impact

No behaviour changes outside `PARALLEL_UPDATES`. Targeted run of `custom_components/haeo/tests/test_init.py` passes (25/25). Ruff and ruff-format clean on the modified Python files.

### DIFF
--- a/custom_components/haeo/entities/haeo_sensor.py
+++ b/custom_components/haeo/entities/haeo_sensor.py
@@ -62,6 +62,12 @@ class HaeoSensor(CoordinatorEntity[HaeoDataUpdateCoordinator], SensorEntity):
         self._attr_unique_id = unique_id
         if translation_placeholders is not None:
             self._attr_translation_placeholders = translation_placeholders
+        # Outputs flagged advanced are LP-internal diagnostics (e.g. battery
+        # aggregate energy flows, SOC envelope bounds) that change at every
+        # optimisation tick. Disable them in the entity registry so the
+        # recorder doesn't track them by default; advanced users can enable
+        # individually if they want to see them.
+        self._attr_entity_registry_enabled_default = not output_data.advanced
         self._apply_output(output_data)
 
         self._record_forecasts = coordinator.config_entry.data.get(CONF_RECORD_FORECASTS, False)
--- a/custom_components/haeo/manifest.json
+++ b/custom_components/haeo/manifest.json
@@ -9,6 +9,7 @@
   "integration_type": "hub",
   "iot_class": "local_push",
   "issue_tracker": "https://github.com/hass-energy/haeo/issues",
+  "quality_scale": "silver",
   "requirements": ["highspy>=1.12.0", "numpy>=1.21.0"],
   "version": "0.4.0rc7"
 }
--- a/custom_components/haeo/manifest.json
+++ b/custom_components/haeo/manifest.json
@@ -9,7 +9,7 @@
   "integration_type": "hub",
   "iot_class": "local_push",
   "issue_tracker": "https://github.com/hass-energy/haeo/issues",
-  "quality_scale": "silver",
+  "quality_scale": "platinum",
   "requirements": ["highspy>=1.12.0", "numpy>=1.21.0"],
   "version": "0.4.0rc7"
 }
--- a/custom_components/haeo/number.py
+++ b/custom_components/haeo/number.py
@@ -29,6 +29,9 @@ from custom_components.haeo.horizon import HorizonManager
 
 _LOGGER = logging.getLogger(__name__)
 
+# Number entities are backed by InputStore state, not by external I/O, so unlimited parallel updates is safe
+PARALLEL_UPDATES = 0
+
 
 def _build_surfaced_mirror_entities(
     config_entry: HaeoConfigEntry,
--- a/custom_components/haeo/quality_scale.yaml
+++ b/custom_components/haeo/quality_scale.yaml
@@ -1,7 +1,7 @@
 version: 1
 integration: haeo
 current_tier: silver
-target_tier: platinum
+target_tier: gold
 
 bronze:
   config-flow:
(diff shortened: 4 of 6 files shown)


purcell-lab:
Pushed `9361540c` to address the Guide tests failure.

### Root cause

`tests/guides/conftest.py` had `pytestmark = [pytest.mark.usefixtures("socket_enabled"), ...]` at module level. **pytest >= 9.0 no longer propagates module-level `pytestmark` from `conftest.py` to collected test items** — only `pytestmark` defined inside a test file (or markers on test classes/functions) applies. So `socket_enabled` was silently a no-op, and every guide test that tried to bind a free port for the live HA HTTP server tripped `pytest_socket.SocketBlockedError` from pytest-homeassistant-custom-component's global socket disable.

This is also why CI started failing on this PR despite the diff being unrelated to tests: the last Guide tests run on main was 2026-06-08 (PR #468), and pytest 9 / the affected behaviour appears to have shipped via uv-sync upgrades in the runner environment between then and now. My PR happens to be the first thing to trigger Guide tests since.

### Fix

Replaced `pytestmark.usefixtures("socket_enabled")` with an explicit autouse fixture in the same conftest:

```python
@pytest.fixture(autouse=True)
def _enable_socket_for_guides() -> None:
    enable_socket()
```

This reliably applies to every test under `tests/guides/` regardless of pytest version.

### Verification

- Reproduced `SocketBlockedError` locally on plain `upstream/main` (`c4f0bfb2`) — confirms not caused by this PR's changes.
- After the fix, the original failure is gone (local run gets past the socket guard to a Playwright firefox-binary check, which only matters in CI where firefox is installed).
- Ruff + ruff format clean.

### Scope-creep note

This is outside the strict "quality-scale housekeeping" scope of the original PR, but it's blocking the merge and the fix is small + self-contained. Happy to split it into a separate PR if you'd prefer @TrentHouliston — let me know.

The `filterwarnings` entries in the remaining `pytestmark` may have the same issue but they weren't the visible failure, so I left them alone for now. If they're also non-functional under pytest 9, a follow-up to move them into the autouse fixture (or onto each test module) would be the cleanest fix.

TrentHouliston (commented):
It's true, I hadn't re-reviewed the quality scale for a while. That said, if the only thing that's making us silver rather than platinum is reviewing the disabled by default for diagnostic sensors. We should just do that and jump to platinum.

Technically these quality scales only exist for integrations that are in core, we will always be quality of "custom". But that doesn't mean we can't aim for the same level that core does.

purcell-lab:
Done — pushed `07b6f688` to take the jump to Platinum.

### What changed

The model already had a per-output `advanced` flag (set by adapters in `core/adapters/elements/`, currently used on four battery aggregate outputs: `BATTERY_ENERGY_IN_FLOW`, `BATTERY_ENERGY_OUT_FLOW`, `BATTERY_SOC_MAX`, `BATTERY_SOC_MIN`). It was already being carried through `CoordinatorOutput` and exposed as an attribute on every sensor, but it wasn't influencing entity registry state.

One-line wiring change in `HaeoSensor.__init__`:

```python
self._attr_entity_registry_enabled_default = not output_data.advanced
```

So the four LP-internal diagnostic battery outputs ship disabled by default. The recorder skips them, users who want them can enable individually. Existing entities aren't affected (entity registry only honours `enabled_by_default` on first creation).

Adding a new diagnostic output type in future is just `advanced=True` on the adapter's `OutputData` — no extra wiring needed.

### `quality_scale.yaml` updates

- `current_tier`: silver → **platinum**
- `target_tier`: gold → **platinum**
- `entity-disabled-by-default`: todo → **done** with the rationale above
- Summary: **41 done / 7 exempt / 0 todo / 100%**
- Added your point about custom-integration badges to the notes:

  > Note: the HA Quality Scale formally applies to core integrations only. HAEO is a custom integration and so its visible badge will remain `quality_scale=custom`, but the implementation meets the same bar as a Platinum core integration.

- `manifest.json` quality_scale: silver → **platinum** (matches yaml; hassfest validates the two are consistent).

### Verification

- `uv run pytest custom_components/haeo/` → **1753 passed, 2 skipped**
- `ruff check` + `ruff format --check` clean on the edited file
- `pyright` strict-mode clean

CI should pick this up shortly. Happy to split the CI conftest fix (`9361540c`) into a separate PR retrospectively if you'd prefer this one to be pure quality-scale work — just say the word.

Issue #253, opened 2026-01-20
Detail reasons for infeasibility


BrendanAnnable:
Currently we just fail at infeasibility, with no context given as to why.

Perhaps we can leverage this: https://www.gams.com/latest/docs/S_HIGHS.html#HIGHS_IIS

Issue #248, opened 2026-01-19
Sporadic erratic infeasible outcomes after upgrading HAEO v0.1.0 to v0.2.1


michiel-dewilde:
### System Health details

N/A

### Checklist

- [x] I have enabled debug logging for my installation.
- [x] I have filled out the issue template to the best of my ability.
- [x] This issue only contains 1 issue (if you have multiple issues, open one issue for each issue).
- [x] This issue is not a duplicate issue of any [previous issues](https://github.com/ludeeus/integration_blueprint/issues?q=is%3Aissue+label%3A%22Bug%22+)..

### Describe the issue

After upgrading HAEO from v0.1.0 to v.0.2.1, my optimization becomes infeasible sporadically.
It happens ostensibly randomly erratically, and not related to the input (when faulty, if I reload HAEO the optimization is an immediate success).

e.g. see this optimization price graph

<img width="1626" height="540" alt="Image" src="https://github.com/user-attachments/assets/b51a42bb-b133-4c49-9233-a4710039bfb4" />

The log message of a failed optimization looks like

```
2026-01-20 00:18:00.604 ERROR (MainThread) [custom_components.haeo.coordinator.coordinator] Unexpected error fetching haeo_01KFC3S2VXRA9E2KPZYDCBF3FN data
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/helpers/update_coordinator.py", line 416, in _async_refresh
    self.data = await self._async_update_data()
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/config/custom_components/haeo/coordinator/coordinator.py", line 565, in _async_update_data
    cost = await self.hass.async_add_executor_job(network.optimize)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.13/concurrent/futures/thread.py", line 59, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/config/custom_components/haeo/model/network.py", line 179, in optimize
    raise ValueError(msg)
ValueError: Optimization failed with status: Infeasible
```

One strange aspect is that I have 2 near-identical HAEO instances, and the issue happens only in one of them:

<img width="1611" height="526" alt="Image" src="https://github.com/user-attachments/assets/3f8fc696-8881-431b-b07f-1a54d127dd01" />

[...]

PR #484, opened 2026-07-12
Recover from infeasible reused optimizer models

barneyonline:
## Summary

Add bounded recovery for the observed persistent infeasibility on reused calibrated optimizer models.

- Raise typed optimization failures instead of parsing generic ValueError messages.
- Clear HiGHS solver state and retry a reused calibrated blended solve once.
- Rebuild once from current loaded participant values and horizon periods when the retry remains infeasible.
- Replace the live network, updater closures, and topology atomically only after the rebuilt solve succeeds.
- Preserve genuine infeasibility when the rebuilt model also fails.

## Related Issues

Related context: [hass-energy/haeo#248](https://github.com/hass-energy/haeo/issues/248) and [hass-energy/haeo#253](https://github.com/hass-energy/haeo/issues/253).

## Type of change

- [x] Bugfix
- [ ] Device support / compatibility
- [ ] New feature
- [ ] Documentation
- [ ] Refactor / tech debt
- [ ] Translation update
- [ ] Other (describe below)

## Testing

```bash
uv run --frozen ruff check .
uv run --frozen pytest -q custom_components/haeo/core/model/tests/test_network.py custom_components/haeo/coordinator/tests/test_coordinator.py
uv run --frozen pytest -q -m "not guide and not scenario"
uv run --frozen pyright custom_components/haeo/core/model/network.py custom_components/haeo/core/model/tests/test_network.py custom_components/haeo/coordinator/coordinator.py custom_components/haeo/coordinator/tests/test_coordinator.py
git diff --check
```

Results:

- Targeted model and coordinator tests: 118 passed.
- Repository non-visual suite: 1,871 passed, 2 skipped, 13 deselected.
- Ruff and Pyright passed.
- Diff validation passed.

## Checklist

- [ ] I updated `CHANGELOG.md` for user-facing changes.
- [ ] I updated documentation (`README.md`, docs/) when behaviour or options changed.
- [ ] I verified translations (`custom_components/haeo/translations/`) are complete and valid.
- [ ] I ran targeted coverage for each touched Python module and confirmed 100% coverage.
- [ ] I reviewed GitHub Actions results (tests, hassfest, quality scale, validate).
- [x] I confirm this PR is scoped to a single logical change set.

## Diagnostics / Screenshots / Notes

The optional guide and scenario visual suites require Playwright Firefox and frontend jsdom assets that are not installed in this workspace. The full non-visual suite passes.


nickw444:
Looks like I bumped into something along these lines today - solution went infeasible. Coding agent indicated the following to me. Followed the suggestion to drop from 5d to 3d forecast horizon and back up and running now:

```
 Entity forecasts are fine. I rebuilt the model from the live HAEO-loaded inputs and the LP is feasible — the
  failure is in HAEO’s default lexicographic/calibrated solve path at your current model size.

  Entity check (live attributes)

  All configured inputs look healthy right now:

  ┌─────────────────────────────────────────────────────┬─────────────────────────────────────────────┐
  │ Input                                               │ Status                                      │
  ├─────────────────────────────────────────────────────┼─────────────────────────────────────────────┤
  │ general_price_pd7_merged / feed_in_price_pd7_merged │ forecast present (365 pts, covers horizon)  │
  ├─────────────────────────────────────────────────────┼─────────────────────────────────────────────┤
  │ historical_load_forecast + FoxESS load              │ OK; fused base load ~1.2–5.0 kW             │
  ├─────────────────────────────────────────────────────┼─────────────────────────────────────────────┤
  │ Solcast day sensors                                 │ detailedForecast present (48×30 min each)   │
  ├─────────────────────────────────────────────────────┼─────────────────────────────────────────────┤
  │ Battery SOC / min SOC                               │ 18% / 5%                                    │
  ├─────────────────────────────────────────────────────┼─────────────────────────────────────────────┤
  │ Power limits                                        │ import 30, export 15, inverters OK          │
  ├─────────────────────────────────────────────────────┼─────────────────────────────────────────────┤
  │ Policy prices                                       │ OK (early-charge negatives are intentional) │
  └─────────────────────────────────────────────────────┴─────────────────────────────────────────────┘

  number.base_load_forecast / number.pv_forecast (what HAEO actually loaded) are aligned: 144 periods, 16:51 → Aug
  2. No NaNs. Supply ≫ fixed load (~2 kW + 0.5 kW Extra Load).

  number.eq4800_charge_cost is unavailable with restored: true — a stale ghost entity, not driving the optimizer.

  What actually fails

  Reproducing with your live numbers:

  • Blended solve: succeeds (cost ≈ $45.7) on the full topology + all 4 policies
  • Calibrated/Lex solve (HAEO default): fails at phase 2 with Error adding constraint to the model
  • Same failure starts around 144 periods / ~6770 vars; at ≤130 periods Lex/Calibrated also succeed

  So this is not “bad forecast attributes” or “load > supply”. It’s HiGHS/highspy choking when HAEO adds the lex
  primary-objective constraint on a 5‑day model. The logged Infeasible is almost certainly a follow-on from that
  failed constraint add on the reused solver.

  Why it can look intermittent

  5_days uses dynamic tier alignment, so period count moves with time of day. When it lands near 144, Calibrated
  breaks; slightly smaller horizons can still work — which matches “worked earlier / broken after restart”.

  Practical workaround

  In HAEO hub options, drop horizon from 5 days → 3 days (or 2 days). That should keep period count under the
  failure threshold and restore Calibrated solves without changing entities.

  Longer-term this needs an HAEO/highspy fix for lex constraint insertion on large models (or a switch to blended
  mode). I can change the horizon in config if you want.
```

barneyonline:
> Looks like I bumped into something along these lines today - solution went infeasible. Coding agent indicated the following to me. Followed the suggestion to drop from 5d to 3d forecast horizon and back up and running now:
> 
> ```
>  Entity forecasts are fine. I rebuilt the model from the live HAEO-loaded inputs and the LP is feasible — the
>   failure is in HAEO’s default lexicographic/calibrated solve path at your current model size.
> 
>   Entity check (live attributes)
> 
>   All configured inputs look healthy right now:
> 
>   ┌─────────────────────────────────────────────────────┬─────────────────────────────────────────────┐
>   │ Input                                               │ Status                                      │
>   ├─────────────────────────────────────────────────────┼─────────────────────────────────────────────┤
>   │ general_price_pd7_merged / feed_in_price_pd7_merged │ forecast present (365 pts, covers horizon)  │
>   ├─────────────────────────────────────────────────────┼─────────────────────────────────────────────┤
>   │ historical_load_forecast + FoxESS load              │ OK; fused base load ~1.2–5.0 kW             │
>   ├─────────────────────────────────────────────────────┼─────────────────────────────────────────────┤
>   │ Solcast day sensors                                 │ detailedForecast present (48×30 min each)   │
>   ├─────────────────────────────────────────────────────┼─────────────────────────────────────────────┤
>   │ Battery SOC / min SOC                               │ 18% / 5%                                    │
>   ├─────────────────────────────────────────────────────┼─────────────────────────────────────────────┤
>   │ Power limits                                        │ import 30, export 15, inverters OK          │
>   ├─────────────────────────────────────────────────────┼─────────────────────────────────────────────┤
>   │ Policy prices                                       │ OK (early-charge negatives are intentional) │
>   └─────────────────────────────────────────────────────┴─────────────────────────────────────────────┘
> 
>   number.base_load_forecast / number.pv_forecast (what HAEO actually loaded) are aligned: 144 periods, 16:51 → Aug
>   2. No NaNs. Supply ≫ fixed load (~2 kW + 0.5 kW Extra Load).
> 
>   number.eq4800_charge_cost is unavailable with restored: true — a stale ghost entity, not driving the optimizer.
> 
>   What actually fails
> 
>   Reproducing with your live numbers:
> 
>   • Blended solve: succeeds (cost ≈ $45.7) on the full topology + all 4 policies
>   • Calibrated/Lex solve (HAEO default): fails at phase 2 with Error adding constraint to the model
>   • Same failure starts around 144 periods / ~6770 vars; at ≤130 periods Lex/Calibrated also succeed
> 
>   So this is not “bad forecast attributes” or “load > supply”. It’s HiGHS/highspy choking when HAEO adds the lex
>   primary-objective constraint on a 5‑day model. The logged Infeasible is almost certainly a follow-on from that
>   failed constraint add on the reused solver.
> 
>   Why it can look intermittent
> 
>   5_days uses dynamic tier alignment, so period count moves with time of day. When it lands near 144, Calibrated
>   breaks; slightly smaller horizons can still work — which matches “worked earlier / broken after restart”.
> 
>   Practical workaround
> 
>   In HAEO hub options, drop horizon from 5 days → 3 days (or 2 days). That should keep period count under the
>   failure threshold and restore Calibrated solves without changing entities.
> 
>   Longer-term this needs an HAEO/highspy fix for lex constraint insertion on large models (or a switch to blended
>   mode). I can change the horizon in config if you want.
> ```

Thanks—this looks related, but it occurs earlier than the reused-model failure addressed here: the initial calibrated solve fails while adding the phase-2 lex constraint. Could you share anonymized diagnostics, the HAEO/HiGHS versions, and the first full “Error adding constraint” log? That would let us reproduce the ~144-period threshold and determine whether this needs a separate checked row-insertion fix.

purcell-lab:
Field report: this PR's exact failure mode hit a production system on 17 Aug 2026, and the recovery here would have shortened it from 55 minutes to one cycle.

Posting the diagnostics @barneyonline [asked for](https://github.com/hass-energy/haeo/pull/484#issuecomment-5115503799), since our incident looks like the same family as [@nickw444's](https://github.com/hass-energy/haeo/pull/484#issuecomment-5103204780).

## Environment

- HAEO `v0.4.1` plus #502 only. Neither this PR nor #485 was present.
- Home Assistant Core 2026.8.2, Supervisor 2026.07.5, OS 18.2.
- `highspy` per manifest is `>=1.12.0`. My reproduction environment resolves 1.12.0. I could not read the version actually installed in the running container, so treat the live version as unconfirmed.
- Solve mode: default calibrated/lex.

## Sequence

| Time (AEST) | Event |
|---|---|
| 05:17:06 | `homeassistant.reload_config_entry` rebuilt the model |
| 05:18:02 to 05:25:08 | 5 x `Exception: Error adding constraint to the model.` raised from highspy `__addRow`, inside `_constrain_objective` at `network.py:500` |
| 05:25 to 06:05 | every solve returns `Infeasible`, 211 occurrences, last at 06:05:37 |
| 06:12:42 | second `reload_config_entry` clears it, outputs live again |

Two stages, and the split maps cleanly onto the two open PRs. Stage 1 is the row-insertion failure that #485 addresses. Stage 2 is the persistent infeasibility on the reused model that this PR addresses. The manual reload that fixed it is the same action this PR automates, so recovery here would have been one cycle rather than 55 minutes without dispatch.

## What we ruled out

Inputs were clean. Because #502 was already deployed, the price forecasts contained zero sub-1e-9 values at incident time, so this was not an input-noise trigger. We also chased an SOC floor theory and discarded it: `min_charge` is a soft floor in HAEO, so an `initial_charge` below it cannot make the LP infeasible. Lowering two floors changed nothing, and only the reload recovered it. The failure was solver state, not inputs.

That also means the tiny coefficient that tripped HiGHS's `small_matrix_value` guard most plausibly arose inside the lexicographic objective row itself rather than arriving from the input side, which is the part #502 structurally cannot reach.

## Relation to @nickw444's report

His trace shows the same thing: calibrated/lex failing at phase 2 with `Error adding constraint to the model`, and his read that the logged `Infeasible` is a follow-on from the failed add on the reused solver. His ~144 period / ~6770 variable threshold is a size trigger, ours was a coefficient magnitude trigger, but both land in the same `_constrain_objective` row insertion and both leave the same orphaned row behind.

One cross-link that may have been missed: #485 was opened on 12 July, about two weeks before the question here about whether this needs "a separate checked row-insertion fix". #485 already is that fix, so these two threads are describing one problem from opposite ends.

## Backport note

I cherry-picked this PR onto `v0.4.1` alongside #485 and it is running in production now. Two things worth flagging for reviewers.

**It conflicts with #485.** Both touch the same recovery path in `coordinator.py`. #485 handles `LexConstraintStateError` inline with its own rebuild and retry, while this PR replaces that block with `_optimize_with_recovery`. I resolved it by keeping this PR's helper and folding #485's exception into it, so both `OptimizationInfeasibleError` and `LexConstraintStateError` take the same recovery:

```python
except (OptimizationInfeasibleError, LexConstraintStateError) as reused_error:
```

Both failures mean the live model cannot be trusted while the inputs may still be satisfiable, so they want identical treatment. This PR's shape is the better base because it swaps the live model in only after the rebuilt one has solved, which the inline version in #485 does not do. That resolution is mine and has not been reviewed by either author, so please treat it as a suggestion rather than a validated merge.

**A caution for anything added around `_ensure_optimal`.** The cold retry tests here stub `run()` while forcing `getModelStatus()` to return `kInfeasible`. That is a state which cannot occur in production, because the status is reported without a solve ever completing. It is fine for this PR as written, but it becomes a hazard for anything that inspects the solver on the failure path.

Concretely, I hit this while adding IIS based diagnostics: `Highs.getIis` segfaults, taking the whole process down rather than raising, when asked to analyse a model that carries rows but has never completed a solve. I verified the boundary across unrun, solved, post `clearSolver`, and clear then run states, and for both feasible and infeasible models, and the crash needs specifically rows present with no completed solve. The guard is to gate on the solver's own report:

```python
if not bool(solver.getInfo().valid):
    return ()  # no completed solve to analyse
```

To be clear, that is a bug in my diagnostics code and not in this PR. I mention it only because these tests are the thing that exposes it, and anyone adding solver introspection on this path will meet it.

PR #485, opened 2026-07-12
Prevent orphaned lexicographic constraints after HiGHS row warnings

barneyonline:
## Summary

Prevent HiGHS row warnings from leaving orphaned lexicographic constraints in reused optimizer models.

The change normalizes tiny nonzero coefficients without changing constraint semantics, validates non-finite inputs, rolls back rows added before `highspy` raises, and rebuilds the network if rollback cannot be trusted. CI now exercises the model suite with both highspy 1.12.0 and 1.15.1.

## Related Issues

- Related to [hass-energy/haeo#484](https://github.com/hass-energy/haeo/pull/484), which handles recovery from infeasible reused blended models. This PR addresses the distinct earlier failure during lexicographic row creation.

## Type of change

- [x] Bugfix
- [ ] Device support / compatibility
- [ ] New feature
- [x] Documentation
- [ ] Refactor / tech debt
- [ ] Translation update
- [ ] Other (describe below)

## Testing

```bash
uv run pytest custom_components/haeo/core/model custom_components/haeo/coordinator/tests/test_coordinator.py -q
uv run --with highspy==1.15.1 pytest custom_components/haeo/core/model -q
uv run pytest -m "not guide and not scenario and not benchmark" -q
uv run pyright
uv run ruff check .
uv run ruff format --check custom_components/haeo/core/model/network.py custom_components/haeo/core/model/tests/test_network.py custom_components/haeo/coordinator/coordinator.py custom_components/haeo/coordinator/tests/test_coordinator.py
uv run lint-imports
git diff --check
```

Results:

- Regular suite: 1,825 passed, 2 skipped, 69 deselected.
- highspy 1.15.1 model suite: 189 passed, 2 skipped.
- Pyright, Ruff, formatting, import contracts, and diff checks passed.

## Checklist

- [ ] I updated `CHANGELOG.md` for user-facing changes.
- [x] I updated documentation (`README.md`, docs/) when behaviour or options changed.
- [ ] I verified translations (`custom_components/enphase_ev/translations/`) are complete and valid.
- [ ] I ran targeted coverage for each touched Python module and confirmed 100% coverage.
- [ ] I reviewed GitHub Actions results (tests, hassfest, quality scale, validate).
- [x] I confirm this PR is scoped to a single logical change set.

## Diagnostics / Screenshots / Notes

The compatibility job pins highspy 1.12.0 and 1.15.1 independently while retaining the existing open dependency range.


purcell-lab:
Field report: stage one of a production wedge on 17 Aug 2026 was exactly this PR's failure mode, and the trigger was a coefficient that no input-side fix could have caught. Two notes below, one supporting the change and one about the branch state.

## The incident

Running HAEO `v0.4.1` plus #502, without this PR or #484. Home Assistant Core 2026.8.2, Supervisor 2026.07.5, OS 18.2. `highspy` per manifest is `>=1.12.0`, my reproduction environment resolves 1.12.0, and I could not read the version installed in the running container, so treat the live version as unconfirmed. Default calibrated/lex solve mode.

| Time (AEST) | Event |
|---|---|
| 05:17:06 | `homeassistant.reload_config_entry` rebuilt the model |
| 05:18:02 to 05:25:08 | 5 x `Exception: Error adding constraint to the model.` raised from highspy `__addRow`, inside `_constrain_objective` at `network.py:500` |
| 05:25 to 06:05 | every solve returns `Infeasible`, 211 occurrences, last at 06:05:37 |
| 06:12:42 | second reload clears it |

The mechanism matches this PR's description line for line. The exception fires part way through the add, so `self._lex_constraint` is never assigned, `_relax_lex_constraint` has no handle to remove the row, and the partially added row survives as a stale `primary <= old_value` bound. Every later phase 1 attempt is then infeasible against a bound from a previous run, which is precisely the 211 Infeasible tail. It persisted for 55 minutes and cleared only on a second reload.

## Why this matters more than the input-side fix

This is the part I think strengthens the case for this PR specifically.

#502 was already deployed on this system, and the price forecasts contained **zero** sub-1e-9 values at incident time. So the input-side snap-to-zero was working and was not bypassed. The system still wedged.

That means the coefficient which tripped HiGHS's `small_matrix_value` guard was most plausibly generated inside the lexicographic objective row itself, during reification of the primary objective as a matrix row, rather than arriving from the input data. If that reading is right, the coefficient normalisation in this PR is doing work that no amount of input sanitising can do, because the offending value never exists in the inputs. I cannot prove the coefficient's origin from logs alone, so I would call that the most likely reading rather than confirmed, but the clean-input observation is solid.

We also excluded the obvious alternative. An SOC floor theory looked promising and was wrong: `min_charge` is a soft floor, so an `initial_charge` below it cannot make the LP infeasible. Lowering two floors changed nothing and only the reload recovered it.

## Cross-link that seems to have been missed

On #484, @barneyonline [asked](https://github.com/hass-energy/haeo/pull/484#issuecomment-5115503799) whether @nickw444's report needed "a separate checked row-insertion fix". This PR, opened 12 July, already is that fix. His report shows the same `_constrain_objective` insertion failing at ~144 periods, a size trigger where ours was a magnitude trigger, both leaving the same orphaned row. The two threads are describing one problem from opposite ends, so it may be worth linking them.

## Branch state

This PR currently reports `mergeable_state: dirty`, so it has picked up conflicts against `main` since the last update on 9 Aug. It will need a rebase before it can merge.

For what it is worth, the cherry-pick onto `v0.4.1` applied with **zero conflicts** and the suite was green, so the change itself is not entangled with recent `main` movement. It is running in production here now alongside #484. The two PRs do conflict with each other in `coordinator.py`, and I have described the resolution I used over on #484.

PR #487, opened 2026-07-13
Ban typing.Any and replace usages with proper typing

TrentHouliston:
## Summary

Follow-on to #466: bans `typing.Any` (and `typing_extensions.Any`) via the same Ruff TID251 banned-api mechanism, so the import is impossible in new code without a justified `# noqa`.

- **~90 files cleaned outright**, preferring precise types over blanket `object`:
  - `PolicyConfigData` + new `is_policy_config_data()` TypeGuard replace `Mapping[str, Any]` through the policy extraction chain
  - `AnyInputFieldInfo` union alias for the constrained `InputFieldInfo` TypeVar (`InputFieldInfo[Any]` → the two concrete instantiations)
  - `SegmentSpec`, `ExpectedOutput`, `Mapping[str, object]` protocol returns, and isinstance narrowing at HA state boundaries
  - **numpy standardized on `NDArray[np.float64]`** — every array in the codebase is created as float64, so the `np.floating[Any]` idiom was never needed; the structured forecast array is `NDArray[np.void]`
- **Structurally-required `Any` stays, justified per file**: dynamic test drivers that monkeypatch flow/element methods, typing reflection (`get_origin`/`get_args`), loose coordinator fixtures with runtime string keys, and the dev simulator's ad-hoc JSON config
- **~82 not-yet-cleaned files are grandfathered** with `# noqa: TID251  # legacy Any usage; migrate to precise types` markers on their imports — a shrink-only set: remove the marker when cleaning a file, never add `Any` to a marked file. `stubs/` is exempt (mirrors external APIs)
- Typing guide gains an **Avoiding Any** section: precise type → TypeVar → `object` + narrowing, with the noqa escape hatch documented
- The tightening surfaced one pre-existing gap, kept as-is and marked: `PolicyRuleData.price` may be `None` for tagging-only rules, but `CompiledPolicyRule` assumes priced rules

## Test plan

- [x] `uv run ruff check` / `uv run ruff format --check`
- [x] `uv run pyright` (strict) — 0 errors
- [x] `uv run pytest` — 1881 passed, 2 skipped, all 6 scenario snapshots pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -10,7 +10,7 @@ from dataclasses import dataclass
 from importlib import import_module
 from logging import config as logging_config_module
 from types import MappingProxyType
-from typing import Any
+from typing import Any  # noqa: TID251  # legacy Any usage; migrate to precise types
 
 from homeassistant.config_entries import ConfigSubentry
 from homeassistant.core import HomeAssistant
--- a/custom_components/haeo/coordinator/coordinator.py
+++ b/custom_components/haeo/coordinator/coordinator.py
@@ -5,7 +5,12 @@ from dataclasses import dataclass
 from datetime import UTC, datetime
 import logging
 import time
-from typing import TYPE_CHECKING, Any, Literal, TypedDict
+from typing import (
+    TYPE_CHECKING,
+    Any,  # noqa: TID251  # legacy Any usage; migrate to precise types
+    Literal,
+    TypedDict,
+)
 
 from homeassistant.components.sensor import SensorDeviceClass, SensorStateClass
 from homeassistant.config_entries import ConfigEntry
--- a/custom_components/haeo/coordinator/coordinator.py
+++ b/custom_components/haeo/coordinator/coordinator.py
@@ -5,12 +5,7 @@ from dataclasses import dataclass
 from datetime import UTC, datetime
 import logging
 import time
-from typing import (
-    TYPE_CHECKING,
-    Any,  # noqa: TID251  # legacy Any usage; migrate to precise types
-    Literal,
-    TypedDict,
-)
+from typing import TYPE_CHECKING, Literal, TypedDict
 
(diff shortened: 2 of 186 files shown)


TrentHouliston:
Follow-up push: a 10-subagent sweep eliminated the grandfathered `legacy Any usage` markers entirely.

- **Any occurrences: 982 → 204** (28 files), and every remaining `Any` import now carries a specific structural justification — HA framework Any-typed signatures, voluptuous schema dicts, `Element`'s invariant output-name parameter, Mock-style dynamic test drivers, and typing reflection. The generic marker set is empty; the typing guide now requires a justification comment for any new noqa.
- JSON payloads use `homeassistant.util.json.JsonValueType` (or a local recursive alias in HA-independent scripts); float coercion uses `_typeshed.ConvertibleToFloat`.
- New shared helpers: `is_number_field_info`/`is_switch_field_info` TypeGuards (class-name based — HA's `frozen_dataclass_compat` shim makes `isinstance` unreliable for entity descriptions), `DecoratorState`/`TagPriceSpec` TypedDicts, and a couple of structural Protocols.

Latent bugs surfaced by honest typing (behavior deliberately unchanged, flagged inline for follow-up):
1. `system_health.py` reads `entry.runtime_data.last_update_success`/`.data`, but `HaeoRuntimeData` only has `.coordinator` — masked today by `ConfigEntry[Any]`.
2. `tools/sim.py --speed` calls `move_to` on the freezegun object instead of the factory returned by `start()`.
3. `uv run diag --file` replay fails with `Connection.__init__() missing 'tags'` — pre-existing, reproduced on the base commit.

Verified: ruff + format clean, pyright strict 0 errors, full suite 1881 passed / 6 scenario snapshots pass.

PR #489, opened 2026-07-19
fix: skip static card URL registration when frontend is not ready

### LINKED ISSUE #488: Boot race: setup crashes when frontend isn't ready yet, taking down default_config
Hi! First off — thanks for HAEO. I've got it running my home battery against Amber Electric's wholesale prices here in Sydney and it's working brilliantly.

I hit a nasty startup bug yesterday that took me hours to untangle, so I wanted to share what happened in case it bites others.

After a normal restart, my whole Home Assistant came up almost empty — only ~27 entities, missing every integration. It turned out HAEO's setup had run before the frontend had finished initialising, and this line in `_async_register_static_frontend_resources` blew up:

```
File "/config/custom_components/haeo/__init__.py", line 102, in _async_register_static_frontend_resources
    add_extra_js_url(hass, url_path)
  File "/usr/src/homeassistant/homeassistant/components/frontend/__init__.py", line 420, in add_extra_js_url
    hass.data[key].add(url)
KeyError: 'frontend_extra_module_url'
```

That exception by itself would have been survivable — but it also cascaded. Right after it, the log shows:

```
[homeassistant.setup] Unable to set up dependencies of 'default_config'. Setup failed for dependencies: ['logbook', 'my']
[homeassistant.setup] Setup failed for 'default_config': Could not setup dependencies: logbook, my
```

… and the boot degraded from there. Because I then restarted a couple more times while figuring it out, the box fell into safe mode, which made the whole thing look much scarier than the actual bug was.

Some details:

- Home Assistant 2026.7.2 on HAOS (amd64), HAEO v0.4.1 installed manually into `custom_components/`
- My `configuration.yaml` does have `frontend:` with `extra_module_url` entries, so the key exists on most boots — this was purely an ordering race; earlier boots with the same config were fine
- Everything else about HAEO (the optimiser itself, sensors, subentries) worked perfectly once the boot completed

What I did locally was wrap the registration loop in a try/except so it logs a warning and moves on:

```python
    for url_path in available_bundles:
        try:
            add_extra_js_url(hass, url_path)
        except KeyError:
            _LOGGER.warning(
                "frontend not ready during setup; skipping HAEO panel resources for %s",
                url_path,
            )
```

With that, every boot is clean. The only downside is the panel resources might not register on an unlucky boot, which seems like a fine trade — the optimiser is the important part.

Happy to open a PR with this if you'd like, or test a proper fix. And again — great project, thanks for building it.

dougrathbone:
Fixes #488

## The bug

On some boots, HAEO's setup crashes with `KeyError: 'frontend_extra_module_url'`. On my system (HA 2026.7.2) that failure then cascades into `default_config` dependency failures (`logbook`, `my`), degrading the whole boot. It's intermittent — the same config boots fine most of the time, which points at a startup ordering race.

## Root cause

`async_setup` calls `_async_register_static_frontend_resources`, which calls `add_extra_js_url(hass, url)` for each card bundle. That helper is effectively `hass.data[DATA_EXTRA_MODULE_URL].add(url)`, and the key is only created when the `frontend` component runs its own setup. If HAEO's setup happens to run first, the lookup raises `KeyError` and the whole integration setup fails.

## The fix

Guard on `DATA_EXTRA_MODULE_URL not in hass.data` before registering the URLs and skip with a warning — the same skip-with-a-log idiom the function already uses for `hass.http` a few lines above. Static path serving still registers (it doesn't depend on frontend state); only the JS URL registration is skipped, and the warning in the log explains why the cards are missing on that boot. The optimizer itself is unaffected, and the cards register normally on the next start.

I went with the upfront guard rather than a try/except around each `add_extra_js_url` call because the key is either present for every bundle or for none of them — one check, one log line, and it matches the existing style of the function.

## The test

`test_async_setup_skips_static_urls_when_frontend_not_ready` removes `DATA_EXTRA_MODULE_URL` from `hass.data` and asserts `async_setup` completes and returns `True` instead of raising. Verified it fails on current main with the exact `KeyError` from the issue. The existing `test_async_setup_registers_static_frontend_resource` covers the normal path where the key is present.

## Checks

- `pytest custom_components/haeo -m "not scenario and not benchmark and not guide"` — 1757 passed, 2 skipped
- `ruff check` / `ruff format --check` — clean
- `pyright` — 0 errors


TrentHouliston:
Thanks for the fix. I've pushed a commit to your branch with some changes (`maintainerCanModify` was on).

The guard stops the crash but converts it into a silent failure: on an unlucky boot the cards are dropped and stay missing until the next restart, with only a log line. For an integration that ships its own Lovelace cards that's harder to debug than the crash.

The race also shouldn't exist. `manifest.json` had:

```json
"after_dependencies": ["http", "recorder"],
```

No `frontend`, despite `async_setup` calling `add_extra_js_url`. `setup.py` schedules and awaits `after_dependencies` before the integration's own setup, so adding `frontend` there makes boot ordering deterministic. That's the root cause.

On top of that, URL registration moves into an `async_when_setup(hass, FRONTEND_DOMAIN, ...)` callback — runs immediately if frontend is already set up, on `EVENT_COMPONENT_LOADED` if not, so the cards register regardless of ordering. `after_dependencies` is soft and gets skipped when frontend isn't scheduled at all, so both parts are load-bearing. Side effect: the integration no longer imports `DATA_EXTRA_MODULE_URL`, which isn't public API of the frontend component. Only the test touches it now.

Your regression test now asserts the URLs arrive rather than that registration was skipped, which lets one test distinguish all three states:

| | result |
|---|---|
| before the guard | `KeyError: 'frontend_extra_module_url'` |
| with the guard | URLs never registered |
| now | URLs registered |

`test_async_setup_registers_static_frontend_resource` also needed a fix: it simulated a ready frontend by planting `hass.data[DATA_EXTRA_MODULE_URL]` without adding `frontend` to `hass.config.components`. Not a reachable state — pre-existing, but invisible while the code read the key directly.

`uv run check` green on all 11 gates locally. Enabling CI now for `hassfest`, since the manifest changed.

dougrathbone:
@TrentHouliston Nice work - glad to have helped in some way, even if I was a little off on root cause.

Thanks for putting this package together!

Issue #37, opened 2025-11-12
Add Deferrable Loads


TrentHouliston:
### Add Support for Deferrable Loads

Haeo does not yet support *deferrable* loads—loads that must consume a certain amount of energy over a day but do not require contiguous operation. This feature would enable optimisation of devices such as dishwashers, EV chargers (in energy-based mode), or other appliances where the *total* daily energy matters more than the exact runtime pattern.

### Proposed Approach

**1. LP-based formulation**
Deferrable loads will be implemented using a linear programming model. Because the daily energy requirement may not align cleanly with the optimisation interval size, the model must support *fractional energy allocation*.

* The decision variable for each interval represents **the fraction (0–1) of that interval during which the load runs**.
* The execution engine interprets this as: *“run the load for X% of this window.”*

**2. Daily energy tracking via sensors**
Inputs will use “total daily energy” sensors to track how much energy the load has consumed so far.

* At the **start of the optimisation day**, the internal energy counter in the model resets to `0`.
* At the **final interval of the day**, the optimisation will enforce that the cumulative energy reaches the configured **daily energy target**.

**3. Slack variable for feasibility**
To avoid infeasible solutions in cases where the device has over- or under-delivered energy (e.g., Home Assistant restarts, missing sensor updates, or unexpected behaviour), introduce a slack variable:

* Accounts for any deviation from the target
* Penalised in the objective function to discourage unnecessary deviation
* Ensures the LP always has a feasible solution, even in edge cases

### Outcome

This feature will allow Haeo to intelligently schedule flexible, energy-budget-based loads while preserving full LP solvability and avoiding infeasible states. It also smoothly integrates with existing Home Assistant energy sensors and the system’s interval-based optimisation framework.

---

principat:
> Hello! Thank you for this project. I really like it.
> 
> Proposal for the EV use case: would it make sense to model the EV as Battery and modify "Minimum charge" to come from a "forecast sensor". This way the charging availability can be controlled with a one-way connection and with forecast sensor we could indicate to the model something like this: "for timesteps 1,2,3 minimum charge is 20% to allow for ad-hoc grocery shopping but at 6AM the minimum charge is 90% so that I can reach my farther away work place". Then the model will charge the battery at optimal times (when the EV is available) to achieve that goal.

[...]

PR #499, opened 2026-08-03
Add the deferrable load as a user-facing element type

TrentHouliston:
:robot: Stacked on #497. Implements the energy-based deferrable load from #37, built on the model element from #494 (the same one the EV's trip sink uses).

## Summary

A new user-facing **Deferrable Load** element for appliances that must absorb a set amount of energy within scheduled windows — pool pumps, hot water, irrigation, or an EV charger driven in energy mode.

## Semantics

- **Run window calendar**: each calendar event is a window; the event text carries the required energy in kWh (location → summary → description, first field that parses as a number). Recurring events give daily/weekly routines.
- **Window gating**: power can only flow to the load while a window is open, capped by the optional **max power**.
- **Priced shortfall** (default \$10/kWh, pre-filled): a window that physically cannot fit its energy prices the deficit instead of breaking the optimization; a missed window stays priced even if a later one catches up. Lower the price to let expensive periods win ("skip the pool clean when energy costs more than it's worth").
- **Optional overage price** for absorbing beyond the requirement.

## Composition

Two model elements: the `deferrable_load` model element (cumulative capacity/requirement profiles from the calendar edges) and one window-masked connection. Sensors: **Power**, **Energy absorbed**, **Energy shortfall** — all with forecast attributes for driving the appliance's switch.

## Testing

- Adapter LP tests: window gating (zero flow outside windows), energy absorbed within the window at the cheapest feasible placement, too-small windows pricing the shortfall, unlimited-power gating, output mapping
- Flow tests: calendar schema value storage, required-calendar validation; parametrized element/translation suites pick the new type up automatically
- `uv run check`: all 11 gates pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/custom_components/haeo/core/adapters/elements/deferrable_load.py
+++ b/custom_components/haeo/core/adapters/elements/deferrable_load.py
@@ -1,0 +1,168 @@
+"""Deferrable load element adapter for model layer integration."""
+
+from collections.abc import Mapping
+from dataclasses import replace
+from typing import Any, Final, Literal
+
+import numpy as np
+from numpy.typing import NDArray
+
+from custom_components.haeo.core.adapters.output_utils import connection_power, expect_output_data
+from custom_components.haeo.core.const import ConnectivityLevel
+from custom_components.haeo.core.model import ModelElementConfig, ModelOutputName, ModelOutputValue
+from custom_components.haeo.core.model.const import OutputType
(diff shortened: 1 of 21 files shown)


TrentHouliston:
:robot: Superseded by #500. The stack was reordered so the deferrable load element sits directly on top of the model element PR (#494) rather than after the EV work — that way it can be reviewed and merged independently of, and before, the EV element. GitHub auto-closed this PR as merged when its former base branch was rewritten; no code was actually merged.

PR #502, opened 2026-08-14
fix(#501): snap sub-1e-9 fused forecast values to zero

### LINKED ISSUE #501: Float-noise sub-1e-9 fused prices trigger HiGHS row warning in lex constraint -> orphaned row -> permanent "Infeasible" (reload/restart do not recover)
# Deterministic solve failure: sub-1e-9 fused prices become HiGHS row warnings -> orphaned lex constraint -> permanent Infeasible

## Summary
A price forecast containing 0 <-> nonzero steps (e.g. Flow Power's happy-hour export tariff:
$0.00 outside a window, $0.35 inside) can wedge HAEO v0.4.1 permanently: every solve fails,
alternating between `ValueError: Optimization failed with status: Infeasible` and
`Exception: Error adding constraint to the model.` (highspy). Config entry reload and full HA
restart do NOT recover while the forecast content stays in the bad regime.

## Mechanism (verified by standalone repro against live data)
1. `_build_extended_block` (forecast_fuser.py) cycle-wraps forecasts with
   `np.mod(timestamp + start_offset, cover_seconds)` on epoch-scale floats (~1.79e9),
   shifting step-edge timestamps by O(1e-7 s).
2. `np.interp` / trapezoidal averaging at a 0 <-> 0.35 step edge then produces values like
   **1.159e-11** instead of exact 0.0 (reproduced at ALL 60 minute-anchors with my live
   Flow Power export forecast, 3_days preset).
3. As phase-1 *costs* these are harmless (phase 1 solves optimally). But `_constrain_objective`
   converts the primary objective into a **matrix row** (`objective <= optimal_value`).
   HiGHS rejects matrix entries with |a_ij| < small_matrix_value (1e-9) with a warning;
   highspy's `addConstr`/`__addRow` raises on any non-kOk status -> "Error adding constraint".
4. The partially-added row is orphaned (`self._lex_constraint` never assigned, so
   `_relax_lex_constraint` can never relax it) -> subsequent phase-1 solves are bound by the
   stale `primary <= old_value` row -> "Infeasible". Repeat attempts add more orphans.
   This matches the intent of PRs #484/#485 but the *input-side* trigger (float-noise tiny
   coefficients) also deserves a fix, since with a standing tariff forecast EVERY rebuild
   (reload/restart) re-fails immediately.

## Observed
- HAEO v0.4.1, HA 2026.7.2, HAOS 17.3. Sensors unavailable from onset; horizon sensor keeps
  ticking; diagnostics endpoint 500s ("no optimization has completed yet").
- Two prior identical incidents (2026-07-27 11:46->21:16 "self-healed" when forecast content
  shifted; 2026-08-09 09:06 onward, survived reload x5 and a full core restart).

## Repro
Standalone harness (no HA): feed a forecast_dict with a 0/0.35 step tariff through
TimeSeriesLoader.load_intervals over a 3_days-preset horizon; scan outputs for
0 < |v| < 1e-9. Hits at every minute anchor.

## Suggested fix
Snap |value| < 1e-9 to 0.0 in `fuse_to_boundaries` / `fuse_to_intervals` (patch attached),
and/or sanitize cost vectors before `_constrain_objective`, and/or tolerate kWarning from
addRow. Happy to PR.

## Patch (running in production since 2026-08-09, immediately resolved a 2h wedge that had survived 5 reloads and a core restart)

```diff
diff --git a/custom_components/haeo/core/data/util/forecast_fuser.py b/custom_components/haeo/core/data/util/forecast_fuser.py
index 5e83da0..5919130 100644
--- a/custom_components/haeo/core/data/util/forecast_fuser.py
+++ b/custom_components/haeo/core/data/util/forecast_fuser.py
@@ -73,6 +73,14 @@ def fuse_to_boundaries(
     # Interpolate at boundary times
     values = np.interp(horizon_times, block_array["timestamp"], block_array["value"])
 
+    # Snap float-noise magnitudes to zero. Cycle-wrapping in _build_extended_block
+    # applies np.mod to epoch-scale timestamps, shifting step edges by ~1e-7 s;
+    # interpolating a 0<->x price step at such an edge yields values like 1e-11.
+    # As objective costs these are harmless, but _constrain_objective turns the
+    # objective into a matrix row, where HiGHS rejects |a_ij| < 1e-9 with a
+    # warning that highspy escalates to "Error adding constraint".
+    values[np.abs(values) < 1e-9] = 0.0
+
     # Replace position 0 with present_value if provided
     result = [float(v) for v in values]
     if present_value is not None:
@@ -137,7 +145,11 @@ def fuse_to_intervals(
 
         # Trapezoidal integration: area under curve divided by duration
         area = np.trapezoid(values, times)
-        result.append(float(area / interval_duration))
+        interval_value = float(area / interval_duration)
+        # Snap float-noise magnitudes to zero (see fuse_to_boundaries).
+        if abs(interval_value) < 1e-9:
+            interval_value = 0.0
+        result.append(interval_value)
 
     # Replace first interval with present_value if provided
     if present_value is not None:
```

purcell-lab:
## Summary

Fixes #501: float-noise sub-1e-9 fused prices trigger a HiGHS row warning in the lex constraint, orphaning the row and wedging the solver into permanent `Infeasible` (reload/restart do not recover).

Snaps `|v| < 1e-9` to exact `0.0` in both `fuse_to_boundaries` (post-`np.interp`) and `fuse_to_intervals` (post-trapezoidal average) so the tiny values never reach the LP as matrix coefficients.

Diagnosis, root-cause analysis, and original patch credit: @kymocode (#501).

## Mechanism (from #501)

1. `_build_extended_block` in `forecast_fuser.py` cycle-wraps forecasts with `np.mod(timestamp + start_offset, cover_seconds)` on epoch-scale floats (~1.79e9), shifting step-edge timestamps by O(1e-7 s).
2. `np.interp` / trapezoidal averaging at a `0 <-> non-zero` step edge (e.g. Flow Power's happy-hour export tariff, `$0.00` outside a window / `$0.35` inside) then produces values like `1.159e-11` instead of exact `0.0`.
3. As phase-1 objective *costs* these are harmless -- phase 1 solves optimally. But `_constrain_objective` reifies the primary objective as a matrix row (`objective <= optimal_value`). HiGHS rejects matrix entries with `|a_ij| < small_matrix_value` (default `1e-9`) with a warning, which highspy escalates on any non-`kOk` status to `Error adding constraint`.
4. The partially-added row is orphaned (`self._lex_constraint` never assigned), so `_relax_lex_constraint` can't remove it, and every subsequent phase-1 solve is infeasible against the stale bound. Config-entry reload and full HA restart do not recover while the forecast content stays in the bad regime.

This is complementary to the constraint-management hardening in #484/#485. Those fix the *aftermath*; this fixes the *input-side trigger*, so the pathological coefficient never enters the model in the first place.

## Change

```python
# fuse_to_boundaries, after np.interp:
values[np.abs(values) < 1e-9] = 0.0

# fuse_to_intervals, after trapezoidal average:
interval_value = float(area / interval_duration)
if abs(interval_value) < 1e-9:
    interval_value = 0.0
```

Threshold `1e-9` matches HiGHS's `small_matrix_value` default -- anything the solver would reject as a matrix entry gets zeroed before it can become one.

## Tests

Two new regressions in `test_forecast_fuser.py`:

- `test_fuse_to_boundaries_snaps_float_noise_to_zero`
- `test_fuse_to_intervals_snaps_float_noise_to_zero`

Both construct a `0 -> 0.35` ramp and place a horizon boundary (or interval sliver) a small fraction of the way up it. Without the snap, `np.interp` / trapezoidal averaging returns `~3.5e-12`; both tests assert no result value is a sub-1e-9 non-zero. Verified to **fail without the patch and pass with it** (the assertion prints the exact `3.5015546018257734e-12` produced without the fix).

Full HAEO suite: **1746 passed, 2 skipped** on `v0.4.1 + patch`.

## Live verification

On a system running v0.4.1, all five HAEO output sensors went `unavailable` simultaneously at `2026-08-15T08:00:14.234608+10:00` and stayed dark for **50 min 12 s** while the coordinator alternated between `Infeasible` and `Error adding constraint`. Pulling `number.grid_export_price` via the HA MCP just before applying the patch showed the pathology in the 192-point forecast attribute exactly as #501 describes:

```text
2026-08-15T08:41:00+10:00  value= 0.0
2026-08-15T08:42:00+10:00  value= 0.0
2026-08-15T08:43:00+10:00  value= 0.0
2026-08-15T08:44:00+10:00  value=-7.549921671549479e-12   <-- #501 float-noise ghost
2026-08-15T08:45:00+10:00  value=-0.0038
2026-08-15T08:46:00+10:00  value=-0.0038
2026-08-15T08:47:00+10:00  value=-0.0038
```

One ghost point at a `0 -> -0.0038` step edge -- three orders of magnitude below HiGHS's `small_matrix_value` (`1e-9`) -- was the sole input-side pathology in this wedge. The `number.grid_import_price` forecast at the same instant contained no sub-1e-9 non-zeros (`min |v| > 0 = 0.0234`), so the export step edge is the only place the bug could have entered the LP.

Applying the patch in Studio Code Server and restarting HA restored solver output within a single solve cycle (first fresh sensor update at `2026-08-15T08:50:25.749779+10:00`). Post-recovery MCP verification of the same forecast attribute:

| Metric | Pre-patch | Post-patch |
|---|---|---|
| forecast points | 192 | 192 |
| exact zeros | 6 | 3 |
| sub-1e-9 non-zeros | **1** (`-7.549921671549479e-12` at 08:44) | **0** |
| min \|v\| > 0 | `7.55e-12` | `0.0026` |
| solver state | wedged | producing output |

All five HAEO output sensors returned to healthy numeric states with 192- or 193-point forecasts, and no `Error adding constraint` / `Infeasible` alternation has been observed since the restart.

## Related

- #484 / #485 -- constraint-management hardening (aftermath fix)
- #501 -- issue with full diagnosis and the original patch by @kymocode

## Base

`v0.4.x` (same track as the current stable line).

### DIFF
--- a/custom_components/haeo/core/data/util/forecast_fuser.py
+++ b/custom_components/haeo/core/data/util/forecast_fuser.py
@@ -73,6 +73,16 @@ def fuse_to_boundaries(
     # Interpolate at boundary times
     values = np.interp(horizon_times, block_array["timestamp"], block_array["value"])
 
+    # Snap float-noise magnitudes to zero (issue #501). Cycle-wrapping in
+    # _build_extended_block applies np.mod to epoch-scale timestamps, shifting
+    # step edges by ~1e-7 s; interpolating a 0<->x price step at such an edge
+    # yields values like 1e-11. As objective costs these are harmless, but
+    # _constrain_objective turns the objective into a matrix row where HiGHS
+    # rejects |a_ij| < small_matrix_value (1e-9) with a warning that highspy
+    # escalates to "Error adding constraint" - orphaning the lex row and
+    # wedging every subsequent solve as Infeasible.
+    values[np.abs(values) < 1e-9] = 0.0
+
     # Replace position 0 with present_value if provided
     result = [float(v) for v in values]
     if present_value is not None:
@@ -137,7 +147,11 @@ def fuse_to_intervals(
 
         # Trapezoidal integration: area under curve divided by duration
         area = np.trapezoid(values, times)
-        result.append(float(area / interval_duration))
+        interval_value = float(area / interval_duration)
+        # Snap float-noise magnitudes to zero (see fuse_to_boundaries, issue #501).
+        if abs(interval_value) < 1e-9:
+            interval_value = 0.0
+        result.append(interval_value)
 
     # Replace first interval with present_value if provided
     if present_value is not None:
--- a/custom_components/haeo/core/data/util/tests/test_forecast_fuser.py
+++ b/custom_components/haeo/core/data/util/tests/test_forecast_fuser.py
@@ -253,3 +253,54 @@ def test_fuse_to_boundaries_raises_when_no_data() -> None:
     """Test that missing both forecast_series and present_value raises ValueError."""
     with pytest.raises(ValueError, match="Either forecast_series or present_value must be provided"):
         fuse_to_boundaries(None, [], [0, 1000, 2000])
+
+
+def test_fuse_to_boundaries_snaps_float_noise_to_zero() -> None:
+    """Regression for issue #501.
+
+    Cycle-wrapping in ``_build_extended_block`` applies ``np.mod`` to
+    epoch-scale timestamps, which shifts step-edge times by ~1e-7 s.
+    Interpolating a 0<->x price step near such an edge produces values like
+    1e-11 instead of exact 0.0. Those tiny values are harmless as objective
+    costs, but ``_constrain_objective`` reifies the objective as a matrix
+    row, and HiGHS rejects coefficients below ``small_matrix_value`` (1e-9),
+    which highspy escalates to ``Error adding constraint`` -- orphaning the
+    lex row and wedging every subsequent solve as Infeasible. The fuser
+    therefore snaps ``|v| < 1e-9`` to exact 0.0 before the value can become
+    an LP coefficient.
+
+    We synthesise the failure by placing a boundary a small fraction of the
+    way up a 0 -> 0.35 ramp so that ``np.interp`` returns a sub-1e-9 value;
+    without the snap, position 1 is a float-noise ghost like 3.5e-12.
+    """
+    forecast_series = [(0, 0.0), (1000, 0.0), (1001, 0.35), (2000, 0.35)]
+    # Boundary sits 1e-11 s past the step start: interp yields ~3.5e-12.
+    horizon_times = [0, 1000 + 1e-11, 2000]
+
+    result = fuse_to_boundaries(None, forecast_series, horizon_times)
+
+    assert result[1] == 0.0, f"expected exact 0.0, got {result[1]!r}"
+    for i, v in enumerate(result):
+        assert v == 0.0 or abs(v) >= 1e-9, (
+            f"position {i}: value {v!r} is a sub-1e-9 nonzero (issue #501 signature)"
+        )
+
+
+def test_fuse_to_intervals_snaps_float_noise_to_zero() -> None:
+    """Regression for issue #501 on the trapezoidal branch.
+
+    Same failure mode as ``fuse_to_boundaries`` -- a narrow interval that
+    barely enters a 0 -> x ramp yields a trapezoidal average below 1e-9 and
+    must be snapped to exact 0.0.
+    """
+    forecast_series = [(0, 0.0), (1000, 0.0), (1001, 0.35), (2000, 0.35)]
+    # Interval [1000, 1000 + 2e-11] straddles a sliver of the ramp; trapezoidal
+    # average lands at ~3.5e-12 without the snap.
+    horizon_times = [0, 1000.0, 1000.0 + 2e-11, 2000]
+
+    result = fuse_to_intervals(None, forecast_series, horizon_times)
+
+    for i, v in enumerate(result):
+        assert v == 0.0 or abs(v) >= 1e-9, (
+            f"interval {i}: value {v!r} is a sub-1e-9 nonzero (issue #501 signature)"
+        )


purcell-lab:
Two updates from running this in production: field evidence on what it does and does not fix, and a lint issue that will fail CI once it runs.

## Field evidence: necessary, but not sufficient

On 17 Aug 2026 this system wedged again, on `v0.4.1` plus this PR. The useful part is what the inputs looked like at the time.

The price forecasts contained **zero** sub-1e-9 values. The snap-to-zero here was working exactly as intended and was not bypassed. Despite that, the solver still wedged:

| Time (AEST) | Event |
|---|---|
| 05:18:02 to 05:25:08 | 5 x `Exception: Error adding constraint to the model.` from highspy `__addRow`, inside `_constrain_objective` |
| 05:25 to 06:05 | 211 x `Infeasible` |
| 06:12:42 | cleared only by `reload_config_entry` |

Same signature as #501, clean inputs. So the residual path is internal: the offending coefficient is most plausibly produced while `_constrain_objective` reifies the primary objective into a matrix row, not carried in from the forecast data.

I read that as good news for this PR's scope. It confirms the fix works at the boundary it claims, and it argues against widening it, because the remaining trigger is not reachable from the input side. The complements are #485 for coefficient normalisation and orphaned-row rollback, and #484 for recovery when a reused model is already poisoned. I have posted the full diagnostics on both.

Worth stating plainly so this PR is not oversold: #501-class input noise is one trigger among at least two, and closing it does not close the wedge.

## Lint: two magic values will fail CI

The project uses `select = ["ALL"]`, and `PLR2004` is only ignored under the test-path overrides, so it is active in `custom_components/`. The two `1e-9` literals in this PR trip it. Verified locally:

```
PLR2004 Magic value used in comparison, consider replacing `1e-9` with a constant variable
   --> custom_components/haeo/core/data/util/forecast_fuser.py:156:34
Found 2 errors.
```

CI has not run yet here, presumably behind the first-time-contributor approval gate, so this has not surfaced as a red check. It will.

The fix is a module-level constant, which also documents why the threshold is what it is:

```python
# HiGHS rejects |a_ij| < small_matrix_value (default 1e-9) as a matrix entry,
# so anything the solver would reject is zeroed before it can become one.
NOISE_FLOOR = 1e-9
```

then `values[np.abs(values) < NOISE_FLOOR] = 0.0` and `if abs(interval_value) < NOISE_FLOOR:`.

I have been running that form in production and `ruff check` is clean with it. Happy to push it to this branch, or leave it if you would rather keep the diff minimal and handle lint at review.

